PR #7, opened 2017-03-19
update mappings with new/changed fields and copy_id for quick search

junjun-zhang:


shanewilson:
@junjun-zhang the gdc_from_graph indexes were not updated:
https://github.com/NCI-GDC/gdc-models/blob/2c8c5f938566da45bee99a79643843a871b29b04/es-models/gdc_from_graph/case.mapping.yaml#L2

PR #9, opened 2017-03-21
add gdc_from_graph setting and mapping files

lindaxiang:


es-models/gdc_from_graph/case.mapping.yaml
```diff
@@ -0,0 +1,2010 @@
...
+                type: keyword
+              relative_with_cancer_history:
+                type: keyword
+              state:
+                type: keyword
+              submitter_id:
+                type: keyword
+              updated_datetime:
+                type: keyword
+          files:
```
junjun-zhang: I am surprised to see another `files` node here, it's like: `case.files.cases.files`. We should only have `case.files.cases`. Checked on `case` document, there is not such fourth level of `files`.

Should remove this in the mapping and try load JSON docs into ES in `strict` mode, if loading passed, it's safe to remove this `files` node.

es-models/gdc_from_graph/file.mapping.yaml
```diff
@@ -0,0 +1,1290 @@
...
+            type: keyword
+          relative_with_cancer_history:
+            type: keyword
+          state:
+            type: keyword
+          submitter_id:
+            type: keyword
+          updated_datetime:
+            type: keyword
+      files:
```
junjun-zhang: Similar to what's commented above in `case` index, this `file.cases.files` node is not need. Should be removed.

es-models/gdc_from_graph/case.mapping.yaml
```diff
@@ -1,3 +1,4 @@
+dynamic: strict
```
junjun-zhang: We don't need `dynamic`

es-models/gdc_from_graph/case.mapping.yaml
```diff
@@ -42,7 +43,7 @@ properties:
         type: keyword
   case_id:
     type: keyword
-    copy_to:
+    copy_to:
```
junjun-zhang: trailing whitespace should be avoided

es-models/gene_centric/gene_centric.mapping.yaml
```diff
@@ -196,6 +196,8 @@ properties:
                         type: keyword
                       dbsnp_rs:
                         type: keyword
+                      dbsnp_val_status:
```
dankolbman: Updates to things not part of the gdc_from_graph should probably be in a separate pr.

junjun-zhang: yes, I agree

lindaxiang: Good point.

PR #21, opened 2017-04-03
chore(observation) [DAT-765]: Add observation_id to gene_centric and …

DrSpez:
…case_centric mappings

### DIFF
--- a/es-models/case_centric/case_centric.mapping.yaml
+++ b/es-models/case_centric/case_centric.mapping.yaml
@@ -377,6 +377,8 @@ properties:
           observation:
             type: nested
             properties:
+              observation_id:
+                type: keyword
               center:
                 type: keyword
               input_bam_file:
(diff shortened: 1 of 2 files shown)


junjun-zhang:
Do we have plan to add `consequence_id` into gene_centric and donor_centric indexes? It's currently in ssm_centric and ssm_occurrence_centric only.

junjun-zhang:
This one will be merged, it's just about timing. The API/UI devs wish to have this ES mapping synchronized with what ES indexes use.

@DrSpez do you know whether the ES indexes to be used by UAT have this `observation_id` field in case_centric and gene_centric? How about `consequence_id` field?

DrSpez:
This change has not made it to the UAT, no worries

DrSpez:
@junjun-zhang yes, we should add `consequence_id` for sure! I will add a new pr for that

PR #29, opened 2017-04-20
Id fields consistency

DrSpez:
- Add `observation_id` and `consequence_id` to all indices for consistency
- Fixed `normal_genotype` mapping


DrSpez:
closed #21 as it is contained in this one

PR #38, opened 2017-06-08
PRTL-1279 mappings for save sets

cy:
I have no idea what I'm doing :D any chance it's remotely correct @junjun-zhang @shanewilson

### DIFF
--- a/es-models/case_set/case_set.mapping.yaml
+++ b/es-models/case_set/case_set.mapping.yaml
@@ -1,0 +1,10 @@
+_all:
+  enabled: false
+properties:
+  case_ids:
+    type: nested
+    properties:
+      case_id:
+        type: keyword
+  case_set_id:
+    type: keyword
--- a/es-models/case_set/case_set.mapping.yaml
+++ b/es-models/case_set/case_set.mapping.yaml
@@ -2,9 +2,4 @@ _all:
   enabled: false
 properties:
   case_ids:
-    type: nested
-    properties:
-      case_id:
-        type: keyword
-  case_set_id:
     type: keyword
--- a/es-models/case_set/case_set.mapping.yaml
+++ b/es-models/case_set/case_set.mapping.yaml
@@ -1,5 +1,5 @@
 _all:
   enabled: false
 properties:
-  case_ids:
+  ids:
     type: keyword
--- a/es-models/case_set/settings.yaml
+++ b/es-models/case_set/settings.yaml
@@ -1,2 +1,1 @@
-index.mapping.nested_fields.limit: 100
 index.max_result_window: 100000000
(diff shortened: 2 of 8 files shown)


es-models/case_set/case_set.mapping.yaml
```diff
@@ -0,0 +1,10 @@
+_all:
+  enabled: false
+properties:
+  case_ids:
```
shanewilson: don't think this needs to be nested - since it is just the one field you can make it `keyword`. ES doesn't care about array types.

es-models/case_set/case_set.mapping.yaml
```diff
@@ -0,0 +1,10 @@
...
+    type: nested
+    properties:
+      case_id:
+        type: keyword
+  case_set_id:
```
shanewilson: can probably get away without this too. just using the `_id` that ES sets.

es-models/case_set/case_set.mapping.yaml
cy: so just
```
case_ids:
  type: keyword
```
and can store a list of ids in there?

es-models/file_set/file_set.mapping.yaml
```diff
@@ -2,9 +2,4 @@ _all:
   enabled: false
 properties:
   file_ids:
```
shanewilson: can we actually change all of these to just `ids`?

cy: sure

es-models/case_set/settings.yaml
```diff
@@ -0,0 +1,2 @@
+index.mapping.nested_fields.limit: 100
```
junjun-zhang: There is no nesting here, don't really need `nested_fields.limit`, ES default will work

es-models/file_set/settings.yaml
```diff
@@ -0,0 +1,2 @@
+index.mapping.nested_fields.limit: 100
```
junjun-zhang: don't really need `nested_fields.limit`

es-models/gene_set/settings.yaml
```diff
@@ -0,0 +1,2 @@
+index.mapping.nested_fields.limit: 100
```
junjun-zhang: don't really need `nested_fields.limit`

es-models/ssm_set/settings.yaml
```diff
@@ -0,0 +1,2 @@
+index.mapping.nested_fields.limit: 100
```
junjun-zhang: don't really need `nested_fields.limit`

cy:
@junjun-zhang @shanewilson changes made - merge ready?

PR #40, opened 2017-07-28
Jbarno/update all by hand

jbarno:
Verified locally that his fixes the esbuild unit test: 

https://travis-ci.com/NCI-GDC/esbuild/builds/51074911#L829

Once this is in I want to tag it with 0.1.1. I'll update the pin in esbuild/dev-requirements with the tag.


es-models/gdc_from_graph/case.mapping.yaml
```diff
@@ -48,10 +48,18 @@ properties:
...
+    type: long
   demographic:
     properties:
+      cause_of_death:
+        type: keyword
       created_datetime:
         type: keyword
+      days_to_birth:
+        type: long
+      days_to_death:
```
junjun-zhang: This field (and some others as well) previously exists under `diagnoses`, it seems like it still exists there, is it intended to duplicate it? Or the old field will be removed eventually?

jbarno: There appear to be a lot of duplicates, I think entire property lists are duplicated between files and cases, I want to look into that further and see if we can make one source of truth for both (I think diagnoses is an example of that problem)

jbarno:
@madopal I have the beginnings of a script to let the output of failing unit tests to be piped back to this. It would be a lot simpler if the fields were sorted lexicographically so I want to do that first in isolation. Then we could add a sorted check to the format_yaml git hook. I don't really know if we ever expect things to be deleted, but there's an issue within this repo that some cases: -> files: -> cases: is unused. So it might be nice to handle the removal process also.

es-models/gdc_from_graph/case.mapping.yaml
dmiller15: We are creating several AB instances in the dictionary with this minirelease. There will be duplicate properties until we successfully migrate the existing fields. Please refer to the individual tickets in the release ticket for the reasoning behind the various additions.

junjun-zhang:
Thanks everyone's comments. There are certainly improvement needed, but the very reason we have this one place to manage Elasticsearch model is to solve a messy dependency graph with model content hard-coded (or partial generated) scattered in multiple components.
In terms of automate update so that it synchronizes with dictionary, it's doable to some degree, we need to approach it carefully. Elasticsearch as it's own unique needs at different places.

es-models/gdc_from_graph/case.mapping.yaml
junjun-zhang: OK, I remember this was discussed. There will be period of time 'mix' old and new will coexist. As brought up earlier, ES indexes built with data in this 'mix' state may not be able to drive the Data Portal in a desirable way. We probably don't need to worry about this now.

Issue #28, opened 2017-04-19
[gdc_from_graph/case] nested files field does not need nested cases


shanewilson:
https://github.com/NCI-GDC/gdc-models/blob/master/es-models/gdc_from_graph/case.mapping.yaml#L517

The structure is 

```
cases:
  files:
    cases: <-- this one is not needed since all case filtering is done on the top level case
```

junjun-zhang:
The original design for such model was to serve the use case to find files that are associated with specific type of biospecimen entities.

Here is an example to demonstrate the use case: give me `cases` who have `aligned reads` files for `tumour` samples.

The query should return `case A` but not `case B`. In order to achieve that the `sample_type` filter has to be applied to `cases.files.case.samples.sample_type`, not `cases.samples.sample_type`.

```
cases:
  - case: A
    samples:
    - sampleA:
        sample_type: normal
    - sampleB:
        sample_type: tumour
    files:
      - fileX:
         data_type: aligned reads
         caseA:
            samples:
            - sampleB:
                 sample_type: tumour
  - case: B
    samples:
    - sampleC:
        sample_type: normal
    - sampleD:
        sample_type: tumour
    files:
      - fileY:
         data_type: aligned reads
         caseB:
            samples:
            - sampleC:
                 sample_type: normal
```

junjun-zhang:
@shanewilson, if we don't need to support this type of queries, then it's OK to remove the second nested `cases`.

philloooo:
closed by #44

PR #44, opened 2017-10-02
Feat/prune cases

DrSpez:
removed `cases[]` from `case.files[].cases[]`


philloooo:
resolve #28 , @NCI-GDC/oicrdevs ?

junjun-zhang:
The change is very straightforward, I don't see any problem. I know currently we don't really have meaningful tests in place, but longer term maybe we should consider add more tests.

DrSpez:
ty @junjun-zhang, merging

PR #46, opened 2017-10-04
nest slides directly under samples

cy:


junjun-zhang:
Do we need it here as well: https://github.com/NCI-GDC/gdc-models/blob/master/es-models/case_centric/case_centric.mapping.yaml ?

cy:
@junjun-zhang updated case_centric, didn't see samples there as well so added the whole samples structure, along with slide_ids and sample_ids (can see them in case_centric/_mapping)

junjun-zhang:
@cy I was just asking, it actually not necessary to have any of the biospcimen related nodes in the 'viz' indexes. This is why the 'viz' indexes don't actually have 'sample' under it.

junjun-zhang:
Close this, duplicated with #47

PR #47, opened 2017-10-04
fix(gdc-models): Add slide underneath sample in tree

madopal:
- Add slide underneath sample in tree


junjun-zhang:
We should not need this addition 'slide'. See comment here: https://jira.opensciencedatacloud.org/browse/SV-842

junjun-zhang:
Discussed with UC team, no need to make any ES mapping change.

PR #52, opened 2017-11-16
fix(path): Fix resource path for packaging

christiandompierre:


junjun-zhang:
Not sure why this is needed. As far as I see it's working fine.

This python module is already imported in `gdcapi`: https://github.com/NCI-GDC/gdcapi/blob/develop/requirements.txt#L38 to pull ES mapping.

Just tested locally, works well for me as well.

junjun-zhang:
Also, it's the intention to not put `es-models` under `gdcmodels`. `gdcmodels` is Python specific, `es-models` is language independent. If one day we need to add wrapper for another language, we don't need to move it around.

christiandompierre:
This is needed in order to be able to build and distribute gdc-models as a python egg. The previous way works when you simply import gdc-models and execute locally, but in order to run the mutation indexer on the spark cluster, we need to build an egg and distribute it and this is the way to do it.

christiandompierre:
as for keeping es-models out of gdcmodels, this doesn't work either when building an egg.

junjun-zhang:
Just built the egg using `python setup.py bdist_egg`. The egg (`dist/gdcmodels-0.1.1-py2.7.egg`) works well.

Any particular error you get?

junjun-zhang:
Just created #53, could it be related to that?

junjun-zhang:
@christiandompierre should I close this?

christiandompierre:
No, don't close this PR. I know how to build an egg, that's not the issue. The issue is that the egg that is built with your version of the gdc-models is broken. you're using `abspath(dirname())` so when the egg is built the absolute path *on the machine the egg was built* is hardcoded. If you distribute this egg to another machine it won't work

junjun-zhang:
OK, in that case, is it possible to fix that part only, but not to move the `es-models` folder?

jbarno:
@christiandompierre you can package data outside of the python package which we do here: [code](https://github.com/NCI-GDC/gdc-models/blob/8c13d564b56ad5afc399d08d4fafdd3a8406a334/setup.py#L26) Note this is following the symlink in the _git_ repo.

If you read between the lines of this [post](https://stackoverflow.com/questions/4519127/setuptools-package-data-folder-location) you can get at where things aren't lining up. Christian went for the Option 1 of the most up-voted solution. Move the data into the package. Same data, just moved down a directory. 

When you generate the egg within the git repository, that symlink in setup.py exists and packaged data is zippedup good to go. When you bring this in as a dependency and call [get_es_models](https://github.com/NCI-GDC/gdc-mutation-indexer/blob/26e7fa6222900423b0fd4b3b2748960ace05c445/exports/mappers/models_mapper.py#L37) this file is actually running inside a virtual environment so doing the absolute path isn't a good idea. The problem Option 1 is alluding to in that stack overflow post. 

I think this might be a good albiet kinda ugly [compromise](https://stackoverflow.com/questions/4519127/setuptools-package-data-folder-location/26533921#26533921)

jbarno:
If this works here's a compromise: https://github.com/NCI-GDC/gdc-models/pull/56/files

PR #55, opened 2017-12-06
is_canonical under gene_centric and case_centric

cy:


junjun-zhang:
Had discussions, we may not need this. Will close once confirmed.

junjun-zhang:
It has been confirmed, we don't need this. The UI will simply provide explanation that only canonical transcript's impact is shown.

PR #56, opened 2017-12-07
Jbarno/kill symlinks

jbarno:
Dunno if this will work with mutation indexer, but its a mix of Christians changes and modifying setup to dynamically pull es-models without having to move it. 

Basically [this](https://stackoverflow.com/questions/4519127/setuptools-package-data-folder-location)

Also it will stop me from consistently doing the symlink loop when looking at this github page :)


DrSpez:
Is it working?

junjun-zhang:
Solution identified in #52. Closing this one.

jbarno:
Actually we need this. Mutation indexer's version of pip really doesn't like the symlinks.

jbarno:
Removing the symlink from the git repo, and moving it to be dynamic in setup.py. Works with older versions of pip.

PR #60, opened 2018-02-02
feat(_size)[DAT-1271]: _size field added

DrSpez:
Adds `_size` field to all doctypes to store document sizes (works with this plugin https://www.elastic.co/guide/en/elasticsearch/plugins/6.1/mapper-size.html)


DrSpez:
Not needed

PR #61, opened 2018-04-26
fix(mapping): Add mappings for graphql subm to pick up new fields

madopal:
- Add mappings so that graphql submission endpoint will properly
pick up & display the new fields we added, changes were only
made to the gdc_from_graph mappings, as those are the ones
that the submission graphql endpoint uses. Once we have data and
want these in esbuild, we'll have to update from that mapping


anmaxvl:
looks deprecated

PR #62, opened 2018-04-30
Removed file_state field

DrSpez:
Relates to https://jira.opensciencedatacloud.org/browse/PRTL-2109


jbarno:
This is for Icabod correct?

DrSpez:
yep, untill then it is in postgres and es

PR #65, opened 2018-08-10
[fix][API-530]: Revert "feat(release_number): added version and release number fields"

anmaxvl:
Reverts NCI-GDC/gdc-models#64


anmaxvl:
want to recreate PR, so that it's linked properly to a ticket

PR #70, opened 2018-09-18
chore(mappings): Added new fields to mappings

christiandompierre:
_size, days_to_birth, days_to_death, days_to_treatment, vital_status

This is needed for Jibboo an addresses this ticket: https://jira.opensciencedatacloud.org/browse/TT-804

### DIFF
--- a/es-models/ssm_centric/ssm_centric.mapping.yaml
+++ b/es-models/ssm_centric/ssm_centric.mapping.yaml
@@ -1,5 +1,10 @@
 _all:
   enabled: false
+_size:
+  enabled: true
+_source:
+  excludes:
+  - __comment__
 properties:
   chromosome:
     type: keyword
@@ -282,6 +287,8 @@ properties:
                 properties:
                   created_datetime:
                     type: date
+                  days_to_treatment:
+                    type: long
                   days_to_treatment_end:
                     type: long
                   days_to_treatment_start:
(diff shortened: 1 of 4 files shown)


es-models/ssm_centric/ssm_centric.mapping.yaml
```diff
@@ -1,5 +1,10 @@
 _all:
   enabled: false
+_size:
+  enabled: true
+_source:
+  excludes:
+  - __comment__
```
junjun-zhang: It's ok to keep this like that, I mean `excludes: __comment__`.

When we build index for DAVE CA for production use we must prune `occurrence.*`, ie, `excludes: occurrence.*`.

es-models/ssm_occurrence_centric/ssm_occurrence_centric.mapping.yaml
```diff
@@ -1,5 +1,10 @@
 _all:
   enabled: false
+_size:
+  enabled: true
+_source:
+  excludes:
+  - __comment__
```
junjun-zhang: similar as the comment in `ssm_centric` mapping, when build index for DAVE CA, pruning needs to be enabled, specific: `excludes: case.observation.*`.

es-models/ssm_centric/ssm_centric.mapping.yaml
christiandompierre: Yeah this PR is not for DAVE CA, this is for Jibboo

PR #73, opened 2018-10-18
Merge upstream changes into release/lorax

singergr:
Merge upstream changes from master into `release/lorax`.

This should address issues with indexing certain cases in Lorax, due to, e.g., `demographic.days_to_birth` missing from `case_centric`.


es-models/gene_centric/gene_centric.mapping.yaml
```diff
@@ -152,6 +154,8 @@ properties:
             properties:
               created_datetime:
                 type: date
+              days_to_treatment:
```
rosibaj: I though days_to_treatment had been removed? I'm not sure if this should still be here - someone else would need to confirm. 
https://jira.opensciencedatacloud.org/browse/TT-417

rosibaj: also this commit: https://github.com/NCI-GDC/gdc-models/commit/0b9490bd0ea5d8daaa5689cc22dfd7cfa038c6f7#diff-c2f9bca69067af7947739862c15f36f9

es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -46,6 +52,12 @@ properties:
         type: keyword
       updated_datetime:
         type: date
+      vital_status:
```
rosibaj: If this is a new field, I think this should just type keyword:
      vital_status:
        type: keyword

es-models/gene_centric/gene_centric.mapping.yaml
singergr: Thanks for the confirmation. We've reviewed the indices that were built for Lorax, and days_to_treatment is not set on any of the documents. I believe the option to filter on days_to_treatment was also removed from the portal some time ago. So, I think the presence of this field at least isn't visible to the end user. We should still remove it from the mapping, but I don't think it is available anywhere in the portal right now.

singergr:
Since we built the current Lorax indices with these mappings and I don't believe we included any values for `days_to_treatment`, I plan to merge this into the Lorax branch later today, and create a follow-up PR against master for the issues noted above. If anyone sees an issue with that plan, please let me know. Thanks!

PR #74, opened 2018-10-22
fix(mappings): fix issues noted in lorax/jibboo

singergr:
Remove `days_to_treatment` from all indices (https://jira.opensciencedatacloud.org/browse/TT-417). Make `vital_status` a keyword rather than a free-text field.

Let's leave Lorax alone since we've already built those indices, and put these changes in master so that we'll pick up the changes in Marvin.


singergr:
Handled in https://github.com/NCI-GDC/gdc-models/pull/78

PR #79, opened 2019-01-15
Feat/tt 928/prune case input files

DrSpez:
Remove `case.files.analysis.input_files`
Corresponds to https://github.com/NCI-GDC/esbuild/pull/220
Do not merge yet. This will `[possibly]` go out after Ned


DrSpez:
Experimental, not needed

PR #85, opened 2019-03-04
feat(sample_type): Adds sample_type to viz indices

christiandompierre:


es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -531,6 +531,12 @@ properties:
         type: boolean
       state:
         type: keyword
+  samples:
+    properties:
+      sample_type:
+        type: keyword
+      sample_type_id:
```
rosibaj: Sample_type looks good here; I checked out the other field in the dictinoary and the data; We do not need sample_type_id so would remove that

es-models/gene_centric/gene_centric.mapping.yaml
```diff
@@ -337,6 +337,12 @@ properties:
             type: boolean
           state:
             type: keyword
+      samples:
+        properties:
+          sample_type:
+            type: keyword
+          sample_type_id:
```
rosibaj: same comment here on sample_type_id

es-models/ssm_centric/ssm_centric.mapping.yaml
```diff
@@ -480,6 +480,12 @@ properties:
                 type: boolean
               state:
                 type: keyword
+          samples:
+            properties:
+              sample_type:
+                type: keyword
+              sample_type_id:
```
rosibaj: same comment here on sample_type_id

es-models/ssm_occurrence_centric/ssm_occurrence_centric.mapping.yaml
```diff
@@ -339,6 +339,12 @@ properties:
             type: boolean
           state:
             type: keyword
+      samples:
+        properties:
+          sample_type:
+            type: keyword
+          sample_type_id:
```
rosibaj: same comment here on sample_type_id

PR #90, opened 2019-04-15
feat(civic): civic clinic annotation

thanh-nguyen-dang:


ZhenyuZ:
@junjun-zhang based on requirement from previous meetings, we have two use cases, and they should have been all supported by this index change
1. add an external link to the CIVIC variant page
2. provide a checkbox to `display CIVIC variant only`
Your suggestions on additional features is great, and I know ICGC portal is doing that.

christiandompierre:
@junjun-zhang we need to be able to search, that's why we added the new fields to all the indices, but the current requirement are only to filter whether or not there's CIViC variant. Your suggestion is great, but it is out of the scope for now. We'll discus internally to see whether we think we should include what you suggest right away or go with what we currently have and improve later

junjun-zhang:
Sounds good to me. It may be done now or later.

If the idea it to expose it as a boolean facet (as @ZhenyuZ described), it might be better to create a field that is typed as boolean, like: `annotated_by_civic` although `civic.gene_id` field could be used to serve the same purpose via Elasticsearch's field exist filter.

es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -329,6 +326,15 @@ properties:
...
             type: keyword
+          clinical_annotations:
+            properties:
+              civic:
+                properties:
+                  gene_id:
+                    type: keyword
+                  variant_id:
+                    type: keyword
+            type: nested
```
singergr: Does `clinical_annotations` need to be of type `nested`? That would only be necessary if we needed to store multiple CIVIC annotations for the same SSM.

es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -2,9 +2,6 @@ _all:
   enabled: false
 _size:
   enabled: true
-_source:
```
singergr: Sorry, just noticed this. I don't think we want to remove these source exclusions.

junjun-zhang:
It was not mentioned that one of the features to be supported is to add a URL link on gene entity page to point to CIVIC. The easiest and correct way to do is to add the CIVIC gene_id to ES index in this part:
https://github.com/NCI-GDC/gdc-models/blob/master/es-models/gene_centric/gene_centric.mapping.yaml#L560-L577

Just like how other existing links to external dbs (NCBI Gene, OMIM etc) work in the portal: https://portal.gdc.cancer.gov/genes/ENSG00000141510

es-models/case_centric/case_centric.mapping.yaml
thanh-nguyen-dang: thanks, I resolved it

singergr: Could you please revert the changes to the `_source`? We will want to continue to exclude the nested gene data from case centric and the nested case data from gene centric.

thanh-nguyen-dang: I reverted it

PR #98, opened 2019-07-08
Feat/dat 2384 save es builds output final mappings in gdc models

jesuspguofc:


gdcmodels/__init__.py
```diff
@@ -2,7 +2,7 @@
 import yaml
 from os import listdir
 import pkg_resources
-from os.path import isfile, join as pj
+from os.path import isfile,dirname, join as pj
```
anmaxvl:
```suggestion
from os.path import isfile, dirname, join as pj
```

gdcmodels/__init__.py
```diff
@@ -11,6 +11,20 @@ def load_yaml(filename):
         return yaml.safe_load(f)
 
 
+def process_ref(name, reference, path):
+    ref = reference['$ref']
```
anmaxvl: this might break for `*_centric` mappings

gdcmodels/__init__.py
```diff
@@ -11,6 +11,20 @@ def load_yaml(filename):
...
+    filename, key = ref.split('#/')
+    filedata = load_yaml(pj(path,filename))
+    return filedata[key] 
+
+
+def load_definitions(filename):
+    import pdb; pdb.set_trace()
```
anmaxvl: 💣

gdcmodels/__init__.py
```diff
@@ -11,6 +11,20 @@ def load_yaml(filename):
         return yaml.safe_load(f)
 
 
+def process_ref(name, reference, path):
+    ref = reference['$ref']
+    filename, key = ref.split('#/')
+    filedata = load_yaml(pj(path,filename))
```
anmaxvl:
```suggestion
    filedata = load_yaml(pj(path, filename))
```

gdcmodels/__init__.py
```diff
@@ -35,7 +49,10 @@ def get_es_models():
                 es_models[es_index][es_type] = {
                         '_mapping': load_yaml(pj(es_model_dir, es_index, f))
                     }
-
+                defs_name = pj(es_model_dir, es_index,es_type+".".join(['_definitions', 'yaml']))
```
anmaxvl:
```suggestion
                defs_name = pj(es_model_dir, es_index, es_type + '_definitions.yaml')
```
don't really see the need of the `join` here

gdcmodels/__init__.py
```diff
@@ -35,7 +49,10 @@ def get_es_models():
...
+                if isfile(defs_name):
+                    definitions = load_definitions(defs_name)
+                    es_models[es_index][es_type]['_mapping'].update(definitions)
```
anmaxvl: if `$ref` is in the file, you'll want to clean it up as well.

gdcmodels/__init__.py
jesuspguofc: my bad lol

gdcmodels/__init__.py
jesuspguofc: simpler, I like it

gdcmodels/__init__.py
jesuspguofc: that's correct thanks

gdcmodels/__init__.py
jesuspguofc: humm not sure I see how. right now it is running through them without a blowing up... maybe we can sync up offline

gdcmodels/__init__.py
jesuspguofc: yup

gdcmodels/__init__.py
jesuspguofc: yup, sorry ahout that

gdcmodels/__init__.py
anmaxvl: oh, you are right, there's that `isfile` check. my bad.

tests/conftest.py
```diff
@@ -119,3 +119,75 @@ def mock_mappings(tmp_path):
...
+def mock_isfile_def(filename):
+    if os.path.basename(filename) == 'gdc_from_graph':
+        return False
+    elif os.path.basename(filename) == 'definitions.yaml':
+        return True
+    return True
+
+
+@pytest.fixture
+def add_def_file(monkeypatch):
```
anmaxvl: can we mock the filesystem instead of mocking calls? similar to what's being done in `mock_mappings` fixture?

jesuspguofc: you got it boss

PR #99, opened 2019-07-08
feat(legacy-mappings): check-in legacy mappings

anmaxvl:
Since we are breaking GDC API's dependency on ESBuild, we need to load mappings for Legacy ES indices from the same source as the other mappings, i.e. here.

### DIFF
--- a/tests/test_es_models.py
+++ b/tests/test_es_models.py
@@ -23,7 +23,7 @@ def test_missing_required_args(input_args):
 def test_get_es_models_standard_behavior():
     models = get_es_models()
 
-    assert len(models) == 11
+    assert len(models) == 12
 
 
 def test_get_es_models_parametrized(mock_listdir):
--- a/es-models/gdc_legacy_graph/annotation.mapping.yaml
+++ b/es-models/gdc_legacy_graph/annotation.mapping.yaml
@@ -1,0 +1,63 @@
+properties:
+  annotation_id:
+    type: keyword
+  case_id:
+    type: keyword
+  case_submitter_id:
+    type: keyword
+  category:
+    type: keyword
+  classification:
+    type: keyword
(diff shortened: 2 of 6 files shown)


tests/test_es_models.py
```diff
@@ -23,7 +26,10 @@ def test_missing_required_args(input_args):
 def test_get_es_models_standard_behavior():
     models = get_es_models()
 
-    assert len(models) == 11
+    models_dir = pkg_resources.resource_filename('gdcmodels', 'es-models')
```
singergr: This is really close to the code being tested. Maybe the hardcoded number is better in some way.

anmaxvl: yeah, makes sense.

tests/test_es_models.py
```diff
@@ -1,3 +1,6 @@
+import os
+import pkg_resources
+
```
scintillating7: I think these are unnecessary

tests/test_es_models.py
```diff
@@ -23,7 +26,7 @@ def test_missing_required_args(input_args):
 def test_get_es_models_standard_behavior():
     models = get_es_models()
 
-    assert len(models) == 11
+    assert len(models) == 12
```
scintillating7: it's a little weird that we have mutation indices unfolded but graph indices folded under 'gdc_from_graph', 'gdc_legacy_graph'.

singergr: It matches the actual index structure, though... we have 4 different doc types stuffed into one graph index.

anmaxvl: as, @singergr mentioned, it's `<index>/<doc_type>.mapping.yaml` structure, it just happened that viz indices have only 1 `doc_type` 🤷‍♂

PR #107, opened 2019-08-08
check in descriptions, update unit tests and some method names

anmaxvl:
Check in field descriptions for both active and legacy graphs:
* dictionary version used for them is `1.18.0`
* update unit tests


gdcmodels/__init__.py
```diff
@@ -14,14 +12,15 @@ def load_yaml(filename):
...
 
-def load_definitions(filename):
-    definitions = load_yaml(filename)
+def load_descriptions(filename):
+    descriptions = load_yaml(filename)
     result = {}
     # we can select what properties to add in the future
-    if '_meta' in definitions.keys():
-        result['_meta'] = definitions.get('_meta')
+    if '_meta' in descriptions:
```
scintillating7: 👍

es-models/gdc_from_graph/descriptions.yaml
```diff
@@ -0,0 +1,3234 @@
+_meta:
+  descriptions:
+    annotations.aliquot.aliquot_quantity: 'The quantity in micrograms (ug) of the
+      aliquot(s) derived from the analyte(s) shipped for sequencing and characterization.
+
```
scintillating7: wondering about these extra newlines...

anmaxvl: yeah, `yaml` was spitting them out like that. did a `strip` before dumping

PR #108, opened 2019-08-09
[FIX] Civic Again

huarngpa:
Did we miss the type as nested?


huarngpa:
> Did we miss the type as nested?

It looks like comments got wiped from the previous [PR](https://github.com/NCI-GDC/gdc-models/pull/106). Maybe this is not needed.

huarngpa:
> > Did we miss the type as nested?
> 
> It looks like comments got wiped from the previous [PR](https://github.com/NCI-GDC/gdc-models/pull/106). Maybe this is not needed.

It looks like there should only be one annotation under ssm @thanh-nguyen-dang

PR #109, opened 2019-08-12
feat(update-mappings): add new mapping properties from gdcdict 1.18.0…

jesuspguofc:
… to fix esbuild errors


jesuspguofc:
No other tests were failing on esbuild, maybe we are not runnig against legacy?
Also, if we are not re-running legacy to create new indices... do those even come into play?

PR #110, opened 2019-08-12
feat(add-normalized-mappings): add normalized mappings for reference

jesuspguofc:


es-models/gdc_from_graph/annotation.mapping.yaml
```diff
@@ -12,68 +12,95 @@ properties:
...
-    copy_to:
-    - annotation_autocomplete
+    copy_to: annotation_autocomplete
+    normalizer: clinical_normalizer
     type: keyword
   case_id:
+    normalizer: clinical_normalizer
     type: keyword
   case_submitter_id:
+    normalizer: clinical_normalizer
```
singergr: Should we also be skipping the normalizer on these fields like `case_submitter_id`, `entity_submitter_id`, etc.?

anmaxvl: good catch, we probably should.

PR #114, opened 2019-09-17
use actual ES mappings dump for legacy

anmaxvl:
Context:
* There is no plan rebuilding legacy graph, but the mappings are out of date at this point, due to change in the dictionary
* We instead use the actual legacy graph mappings directly from Elasticsearch
* This should resolve various issues with missing data for the facets on the legacy archive portal


anmaxvl:
> I think the `descriptions.yaml` is still out of sync. Is that going to be an issue for the API?
> 
> The actual mappings line up, so 👍 for fixing the main issue.

hm. good point actually.

PR #133, opened 2020-05-13
DEV-172 update for elasticsearch 7

singergr:
Update mappings to work with Elasticsearch 7. Some of these changes are backwards-incompatible, so this also updates the mappings to _stop_ working with ES5.

Remove the explicit setting to disable the `_all` field. ES7 no longer supports `_all` at all, so we can't even say to disable it.

Set `index.mapping.nested_objects.limit` to a ridiculously huge value. This is a new setting in ES7 that would otherwise prevent us from storing the tens of thousands of nested documents we sometimes need.

Make setuptools_scm stop complaining if something else tries to pip install this library with `-e` (this was already done on the `release/2.2.0` branch and will be rebased out later).


gdcmodels/init_index.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
```
singergr, line 1: Git couldn't tell that I moved this. 😞

tests/test_es_models.py
```diff
@@ -1,28 +1,6 @@
...
-
-@pytest.mark.parametrize('input_args', [
-    ('--prefix', 'foo-bar'),
-    ('--prefix', 'foo-bar', '--host', 'localhost'),
-    ('--prefix', 'foo-bar', '--index', 'case_centric'),
-    ('--index', 'case_centric gene_centric'),
-    ('--index', 'case_centric', '--host', 'localhost'),
-    ('--host', 'localhost'),
-])
-def test_missing_required_args(input_args):
```
singergr, line 18: Moved to `test_get_parser__requires_args` in the other test file.

singergr:
I think this is ready for review now, but unfortunately prolly needs a re-review.

singergr:
Closing in favor of https://github.com/NCI-GDC/gdc-models/pull/134 since I filed a separate ticket for gdc-models/normalizer.

PR #137, opened 2020-06-09
DEV-213 set huge nested objects limit for legacy

singergr:
Use the same huge value for `index.mapping.nested_objects.limit` in the legacy index that we do in the other indices. Make it possible to populate the legacy graph index on ES7.

### DIFF
--- a/es-models/gdc_legacy_graph/settings.yaml
+++ b/es-models/gdc_legacy_graph/settings.yaml
@@ -10,5 +10,6 @@ analysis:
       min_gram: 2
       side: front
       type: edge_ngram
+index.mapping.nested_objects.limit: 100000000
 index.max_result_window: 100000000
 mapping.nested_fields.limit: 150


es-models/gdc_legacy_graph/settings.yaml
```diff
@@ -10,5 +10,6 @@ analysis:
       min_gram: 2
       side: front
       type: edge_ngram
+index.mapping.nested_objects.limit: 100000000
```
kulgan, line 13: these are really large numbers, won't it affect performance?

singergr: This sets the upper limit on the number of nested documents that may be indexed. I don't believe this would have any impact on how Elasticsearch performs if documents are indexed that don't match that limit.

Our indices already exceed the default limit by quite a bit, and we don't have a way to avoid exceeding the limit without redesigning our index structure.

singergr: I grepped through the ES source to confirm that it doesn't try to use this to size array allocations anywhere (since we have seen size fields do surprising things in the past), and I think we are safe. It looks like it's only used for validation.

https://github.com/elastic/elasticsearch/blob/v7.7.1/server/src/main/java/org/elasticsearch/index/mapper/MapperService.java#L110-L111

https://github.com/elastic/elasticsearch/blob/v7.7.1/server/src/main/java/org/elasticsearch/index/mapper/ParseContext.java#L332

https://github.com/elastic/elasticsearch/blob/v7.7.1/server/src/main/java/org/elasticsearch/index/mapper/ParseContext.java#L376-L385

PR #140, opened 2020-07-09
sync develop -> epic/dave-ca-phase-2

singergr:
Bring epic branch up-to-date with develop.

PR #141, opened 2020-07-20
Sync develop to DAVE-CA Phase 2

singergr:
Replaces https://github.com/NCI-GDC/gdc-models/pull/140.


singergr:
ok merging before this gets away from me again...

PR #147, opened 2020-08-06
Feat/dev 206 gene expression poc

anmaxvl:


.travis.yml
```diff
@@ -1,5 +1,8 @@
 sudo: true
-language: generic
+language: python
+python:
+  - "3.5"
```
singergr, line 4: We should probably keep Python 2.7 here until we've upgraded at least mutation indexer, possible also datatools.

anmaxvl: our jenkins has it and I wanted to drop it here. but yeah, makes sense.

es-models/gene_expression/gene_expression.mapping.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: keyword
+  genes:
+    properties:
+      expression_value:
+        type: float
+      gene_id:
+        type: keyword
+    type: nested
+  hgnc_symbol:
+    type: keyword
```
philtom-ctds, line 23: The `hgnc_symbol` should be a sibling of `gene_id` in the nested gene doc.

tox.ini
```diff
@@ -1,14 +1,14 @@
...
 skip_missing_interpreters=true
 
 [testenv]
 deps =
     -rdev-requirements.txt
 usedevelop=true
 commands =
-    pytest -vvs --cov=gdcmodels --cov-report xml --cov-report html --junit-xml test-reports/results.xml  {posargs}
     python ./scripts/git-hooks/pre-commit
+    pytest -vvs --cov=gdcmodels --cov-report xml --cov-report html --junit-xml test-reports/results.xml  {posargs}
```
singergr, line 11: I agree that it's better to do this second.

We should move the YAML sorting into the pre-commit framework and then run `pre-commit run -a` here instead, so that we apply all of the checks. I can't remember if I already filed a ticket to make a YAML sorter that works with pre-commit. Anyway, that's way outside the scope of this ticket.

es-models/gene_expression/gene_expression.mapping.yaml
anmaxvl: oh lol. good catch

es-models/gene_expression/gene_expression.mapping.yaml
```diff
@@ -0,0 +1,31 @@
+_source:
+  excludes:
+  - genes.*
+properties:
+  age_at_diagnosis:
```
singergr, line 5: I'm not sure which properties we'll really need, but I don't see any reason not to start with this set.

es-models/gene_expression/gene_expression.mapping.yaml
```diff
@@ -0,0 +1,31 @@
+_source:
```
singergr, line 1: It might be interesting to also enable `_size`, but I don't think it matters much either way. This index won't have a document size explosion.

es-models/gene_expression/gene_expression.mapping.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: keyword
+  gender:
+    type: keyword
+  genes:
+    properties:
+      expression_value:
+        type: float
+      gene_id:
+        type: keyword
+      hgnc_symbol:
```
singergr, line 21: I think this field is just called `symbol` in gene_centric, and mutation indexer maps it from the other name in the MAF. We might want to use the same name... probably not a big deal...

singergr: ...or maybe it's `hgnc`? Now I'm not actually sure.

singergr: ...okay, I looked more. I think `symbol` in gene_centric is the HGNC symbol, and `hgnc` is the "HGNC ID," which is something else. So `symbol` is probably a more consistent name.

anmaxvl: I'm following agreement between @ZhenyuZ and @philtom-ctds from a slack conversation they had a couple of days ago

es-models/gene_expression/gene_expression.mapping.yaml
philtom-ctds: I believe these are the ones that oncogrid uses.  So, that's what everyone asked for.

tox.ini
anmaxvl: maybe we can use one from (here)[https://pre-commit.com/hooks.html]?

es-models/gene_expression/gene_expression.mapping.yaml
anmaxvl: discussed offline and decided to use `symbol`

PR #149, opened 2020-08-07
DEV-234 add script to force merge es indices

qiaouchicago:


gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument("--user", default="", help="Elasticsearch client user")
+    parser.add_argument(
+        "--password", dest="password", default="", help="Elasticsearch client password"
+    )
+
+    return parser
+
+
+def _force_merge_wrapper(args):
+    es = Elasticsearch(
```
anmaxvl, line 48: you can probably reuse [this](https://github.com/NCI-GDC/gdc-models/blob/develop/gdcmodels/init_index.py#L80-L86)

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+        hosts=[{"host": args.host, "port": args.port}],
+        use_ssl=True,
+        verify_certs=args.verify_certs,
+        http_auth=(args.user, args.password),
+    )
+
+    return force_merge_elasticsearch_segments(es, args.index)
+
+
+def _force_merge(es, index):
```
anmaxvl, line 58: I don't think we should allow for `index` to be an empty string or `None`. This should result in a validation error

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+        index = ""
+    elif isinstance(index, str) and index != "":
+        index = [index]
+
+    try:
+        es.indices.forcemerge(index, max_num_segments=1)
+    except ConnectionTimeout:
+        active_count = 1
+        while active_count != 0:
+            res = es.nodes.stats(metric="thread_pool")
```
anmaxvl, line 69: this will be making too many requests, since there's no `sleep` in here.

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+            active_count = sum(
+                stat["thread_pool"]["force_merge"]["active"]
+                for stat in res["nodes"].values()
+            )
+
+
+def force_merge_elasticsearch_segments(es, index=None):
+
+    start = time()
+    t = threading.Thread(target=_force_merge, args=(es, index))
```
anmaxvl, line 79: I don't think you need to use threads here at all. Instead of doing `whilet.is_alive()` you could, instead just check the `active_count` value and use sleep inside `_force_merge`. This way you can get rid of some wrapper functions too

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+
+    print("Starting merging segments.")
```
anmaxvl, line 81: I think it's better to let a logger handle this, instead of just printing it to stdout

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+    t.start()
+
+    spinning_cursor = iter(cycle("|/-\\"))
+    while t.is_alive():
+        print("\r{} still merging.".format(next(spinning_cursor)), end="")
```
anmaxvl, line 86: it probably looks nice, but I don't think we really need this. maybe logging time elapsed and number of active threads will be more informative

gdcmodels/force_merge_segements.py
```diff
@@ -0,0 +1,98 @@
...
+from time import sleep, time
+
+from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import ConnectionTimeout
+
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+def get_parser():
```
anmaxvl, line 19: also, consider separating parser logic from the "service" logic

gdcmodels/force_merge_segements.py
qiaouchicago: Without use_ssl=True, the request will fail for our es7. I guess previously we were using port 443 and this time we changed it to 9211?

Do you want me to modify the function in the init_index and add the flag?

gdcmodels/force_merge_segements.py
qiaouchicago: How does the script decide when to merge on all indices?

https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.client.IndicesClient.forcemerge
> `forcemerge(index=None, params=None, headers=None)`

> index – A comma-separated list of index names; use _all or empty string to perform the operation on all indices

gdcmodels/force_merge_segements.py
anmaxvl: yes. you can make a generic enough method that'll return you an ES client and if you want to keep the old code working as it was, you can add a couple of params (including `use_ssl=True` and `port`) and have some defaults. but my guess would be that this needs to change anyway, because of recent xpack introduction

gdcmodels/force_merge_segements.py
qiaouchicago: Sorry, I do not get it. Could you please explain a little bit more?

gdcmodels/force_merge_segements.py
anmaxvl: I don't think a default to all is a good default. I'm proposing a contract where you MUST specify a set of indices

gdcmodels/force_merge_segements.py
qiaouchicago: Forgot to add it when I separate thread 😞

gdcmodels/services/force_merge_service.py
```diff
@@ -0,0 +1,28 @@
...
+    except ConnectionTimeout:
+        active_count = 1
+        while active_count != 0:
+            res = es.nodes.stats(metric="thread_pool")
+            active_count = sum(
+                stat["thread_pool"]["force_merge"]["active"]
+                for stat in res["nodes"].values()
+            )
+            logging.info("Still merging. Active thread count: {}".format(active_count))
+            sleep(3)
```
anmaxvl, line 27: I think it's too frequent. something like 10 or 30 secs should be fine. Also maybe define a constant.

tests/test_force_merge_service.py
```diff
@@ -0,0 +1,39 @@
...
+    counts = [1, 0]
+    value = [
+        {"nodes": {"test_node": {"thread_pool": {"force_merge": {"active": count}}}}}
+        for count in counts
+    ]
+    es.nodes.stats = Mock(side_effect=value)
+    force_merge_elasticsearch_indices(es, index)
+    es.indices.forcemerge.asssert_called_once_with(index, 1)
+    es.nodes.stats.assert_called_with(metric="thread_pool")
+    assert es.nodes.stats.call_count == 2
```
philtom-ctds, line 39: I really discourage this style of mocking in unit tests.

The implementation of the test is tightly coupled to the implementation of the function.   This means that we could change the implementation of the function while maintaining the same behavior, but we'd still have to update the unit test. This means that the unit test is not testing behavior.  Instead, it's testing how the function is implemented.

Also, this test is going to incur the sleep time, which will make the tests slow.  We should avoid this in unit tests.

I recommend omitting these two tests altogether since neither of them actually verify that a force merge happened.  Instead, test the script by hand to verify that this function works as intended.

singergr: I was going to say that we could set up integration tests to verify this, and try to adjust the ES settings to encourage a proliferation of segments, but I'm not sure that'd be worth the effort, and we'll need to hand-test the script to verify it anyway.

anmaxvl: this is something I suggested. are you saying that unit tests supposed to verify that the actual force merge happened? I don't think that's the purpose of the unit tests? Making sure that the code behaves as expected based on a different behavior of the external service is what we're testing here. If (for example) postgres changed its query implementation and we had it mocked with a different behavior, we'd have to update our tests in that case too, so I don't see a problem with this.

gdcmodels/esutils.py
```diff
@@ -0,0 +1,42 @@
...
+
+from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import ConnectionTimeout
+
+UPDATE_INTERVAL = 10
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+def get_elasticsearch(args, use_ssl=False, verify_certs=True):
```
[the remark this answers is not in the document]
anmaxvl: this was also something that I suggested to limit the scope of this PR. the intent was to keep the old behavior and make sure the new one works. Fixing the other scripts is outside of this PR

singergr:
> I am leaning towards moving this scripts to data-tools, the elasticsearch instance creation is already handled there and would make sense to keep it consistent. Also having scripts that do different things scattered around might make it harder to recall which repo has which script that does what.

This is more of a shared library than a script. The idea is to have mutation indexer execute it automatically after building indices (https://github.com/NCI-GDC/gdc-mutation-indexer/pull/288), and make it possible for esbuild to reuse it.

In the future, we should probably move this and other ES-related functionality (like the index init stuff) into a separate "ES utils" repo so that it's again easier to keep track of it all, but adding the code to gdc-models gave us a short-term solution given that we will want that mutation indexer PR for DR 26.

kulgan:
> > I am leaning towards moving this scripts to data-tools, the elasticsearch instance creation is already handled there and would make sense to keep it consistent. Also having scripts that do different things scattered around might make it harder to recall which repo has which script that does what.
> 
> This is more of a shared library than a script. The idea is to have mutation indexer execute it automatically after building indices ([NCI-GDC/gdc-mutation-indexer#288](https://github.com/NCI-GDC/gdc-mutation-indexer/pull/288)), and make it possible for esbuild to reuse it.
> 
> In the future, we should probably move this and other ES-related functionality (like the index init stuff) into a separate "ES utils" repo so that it's again easier to keep track of it all, but adding the code to gdc-models gave us a short-term solution given that we will want that mutation indexer PR for DR 26.

There is the library aspect (exposed functions and methods and all) and there is the use of the library functions as a script within this code base (which is what I was referring to), I believe mutation indexer will use this as a library and not call the scripts included here.

gdcmodels/force_merge_indices.py
```diff
@@ -0,0 +1,49 @@
...
+
+FORMAT = "[%(asctime)s][%(name)14s][%(levelname)7s] %(message)s"
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        description="Force a merge on the shards of one or more ES indices"
+    )
+    parser.add_argument(
+        "--verify-certs",
```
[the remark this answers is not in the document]
qiaouchicago: Should I add `ca-certs` or should I remove `verify-certs`

gdcmodels/esutils.py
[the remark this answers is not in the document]
qiaouchicago: If not using `kwargs`, Should I add `ca-certs` or should I remove `verify-certs`? I think adding `ca_certs`, `client_cert` and `client_key` would be too many?

singergr:
> There is the library aspect (exposed functions and methods and all) and there is the use of the library functions as a script within this code base (which is what I was referring to), I believe mutation indexer will use this as a library and not call the scripts included here.

Okay, I see. I... think I'd still lean toward having that script here since that'll ultimately result in it getting deployed to the ES VMs, but I could see it making more sense in data-tools. We could use a general tool refactoring...

gdcmodels/esutils.py
```diff
@@ -0,0 +1,42 @@
...
+    return Elasticsearch(
+        hosts=[{"host": args.host, "port": args.port}],
+        http_auth=(args.user, args.password),
+        use_ssl=use_ssl,
+        verify_certs=verify_certs,
+        timeout=60,
+    )
+
+
+def force_merge_elasticsearch_indices(es, index, max_num_segments=1):
```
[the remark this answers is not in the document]
kulgan: We should be specific and limit to only `forcemerge` parameters we support, I know the use `kwargs` has its upsides like you mentioned but I generally prefer using it when absolutely necessary as it is no longer obvious looking at the code and docstring what is what.

gdcmodels/esutils.py
```diff
@@ -0,0 +1,64 @@
...
+
+from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import ConnectionTimeout
+
+UPDATE_INTERVAL = 10
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+def get_elasticsearch(host, port=9200, user='', password='', use_ssl=True):
```
kulgan, line 14: Since we are pushing towards the library route, it might make sense to not assume how (what extra options are needed etc) the library users wants to connect to elastic-search and accept an appropriate es instance (as provided in this repo by config.es) or provide a much more robust/generic configuration option, like connecting to multiple nodes, authentication, ssl_context etc. I will vote for not exposing this as part of esutil and let users pass it to the exposed library functions. Since there is a script here also, you can make this more specific for that script by moving it to the script module

gdcmodels/esutils.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Args:
+        es: Elasticsearch object, Elasticsearch low-level client
+        index: list of index names to merge
+        max_num_segments: The number of segments the index should be merged into
+
+    Returns:
+
+    """
+    if not index or not isinstance(index, list):
```
kulgan, line 47: The [docs](https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.client.IndicesClient.forcemerge) says the parameter `index` is a comma separated list of index names :thinking:

singergr: This is an unfortunately common problem with the documentation for the Elasticsearch Python client library. The _REST API_ expects a comma-separated string, so the documentation asks for a string, but really, the Python library will take an actual list. The list will be ["escaped"](https://github.com/elastic/elasticsearch-py/blob/7.7.1/elasticsearch/client/utils.py#L12-L20) into a comma-separated string before it is sent to the REST API.

Given that mutation indexer will need to force-merge several indices, I think it would be helpful if this function could accept a list like it did before @qiaouchicago's most recent commit. We should be able to pass that list directly into the client library. We could also make it accept either a list or a string, but our immediate use cases will all involve multiple indices, so the list will be more useful for us.

philtom-ctds:
@qiaouchicago See this comment on your other PR from this morning:
https://github.com/NCI-GDC/gdc-mutation-indexer/pull/288#discussion_r468611779

My comments around get_elasticsearch and all TLS-related issues no longer matter since it's not being re-used as a general purpose library function.

gdcmodels/init_index.py
```diff
@@ -59,6 +59,25 @@ def format_index_name(prefix, index, index_type=None):
...
+    Returns:
+
+    """
+    return Elasticsearch(
+        hosts=[{"host": args.host, "port": args.port}],
+        http_auth=(args.user, args.password),
+        use_ssl=use_ssl,
+        verify_certs=False,
+        timeout=60,
+    )
```
singergr, line 78: FYI we should merge the changes from the Wocket branch so that we can more easily set the proper TLS settings. We should not be setting `verify_certs=False`. We should instead be providing a path to an appropriate CA bundle.

PR #152, opened 2020-08-27
DEV-327 update gene expression settings

philtom-ctds:
This PR contains two commits.  The first one migrates the old git commit hook for formatting yaml into a local pre-commit hook.  This makes it a little more friendly.

The second commit is the real content of this pr.  It updates a setting for the gene expression index. 

See the commit messages for more details.


es-models/gene_expression/settings.yaml
```diff
@@ -22,4 +22,5 @@ analysis:
       type: edge_ngram
 index.mapping.nested_fields.limit: 100
 index.mapping.nested_objects.limit: 100000000
+index.max_inner_result_window: 20000
```
anmaxvl, line 25: maybe just make it 100K in case we want to include more genes later?

philtom-ctds: From es docs:
>  Inner hits and top hits aggregation take heap memory and time proportional to from + size and this limits that memory.

Since this directly affects heap consumption, I purposely limited this to our current max.  If we have a use case in the future where this needs to be increased significantly, we'll likely need to evaluate an appropriate way to limit resource consumption for that use case.

anmaxvl: we pass in an explicit `"size"`? not sure why this setting would affect the heap size. size for inner hits also defaults to something like `10`? anyways.

PR #153, opened 2020-08-30
merge develop to master

philtom-ctds:
Merging develop to master in order to create a version tag for the new gene expression index.


philtom-ctds:
Using the `2.5.0-rc.2` tag in develop instead of merging/releasing.

PR #155, opened 2020-09-16
DEV-382: add clinical fields with data to viz mappings

anmaxvl:
Syncing graph and viz mappings with fields that have values as of DR26.


philtom-ctds (approved):

kulgan (commented):
So I looked at the ticket for this PR and what I understand is that these fields now have data so we need to include them, my question is, is this the basis for updating what shows up in the mappings or there is some context some where that I am totally missing

kulgan (approved):

PR #160, opened 2020-10-15
DEV-418 convert number fields to doubles in viz index mappings

qiaouchicago:


es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -265,23 +265,23 @@ properties:
...
-        type: long
+        type: double
       alcohol_history:
         type: keyword
       alcohol_intensity:
         type: keyword
       asbestos_exposure:
         type: keyword
       bmi:
-        type: float
```
singergr, line 276: I'm surprised `bmi` was already a float here. I thought I remember seeing it quantized in some contexts.

The change should be good, though. I guess we'll see if we still have issues with values being quantized/truncated after we re-run mutation indexer with the new mappings.

singergr (approved):
I skimmed the remaining `long` fields in the viz indices and they also appear to be longs in the graph index, so I think this is good.

philtom-ctds (approved):

PR #162, opened 2020-11-03
DEV-272 update mappings to include new clinical data

qiaouchicago:


kulgan (approved):
lgtm, are we adding this to the viz indices too?

es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -176,6 +176,85 @@ properties:
...
+          bone_marrow_malignant_cells:
+            type: keyword
+          breslow_thickness:
+            type: double
+          circumferential_resection_margin:
+            type: double
+          columnar_mucosa_present:
+            type: keyword
+          created_datetime:
+            type: keyword
```
singergr, line 194: I'm sorry, I forgot one thing. 😞 Could you remove the `created_datetime` and `updated_datetime` fields from the `_centric` mappings _only_? Mutation indexer excludes those two fields when copying data from the graph index: https://github.com/NCI-GDC/gdc-mutation-indexer/blob/master/config.py#L147-L148

Other than that, I think this is good to go.

singergr (approved):

philtom-ctds (approved):

PR #164, opened 2020-12-23
DEV-500 increase max_terms_counts for saved sets

ProfOak:
- add extra test for that setting

### DIFF
--- a/es-models/gdc_from_graph/settings.yaml
+++ b/es-models/gdc_from_graph/settings.yaml
@@ -30,3 +30,4 @@ index.mapping.nested_fields.limit: 100
 index.mapping.nested_objects.limit: 100000000
 index.mapping.total_fields.limit: 1500
 index.max_result_window: 100000000
+index.max_terms_count: 100000
--- a/tests/test_init_index.py
+++ b/tests/test_init_index.py
@@ -72,6 +72,10 @@ def validate_index(es):
             == expected_settings["index.max_result_window"]
         )
 
+        max_terms_count = actual_settings.get("max_terms_count")
+        if max_terms_count:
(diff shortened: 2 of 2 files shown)


es-models/case_set/settings.yaml
```diff
@@ -1 +1,2 @@
 index.max_result_window: 100000000
+index.max_terms_count: 100000
```
philtom-ctds, line 2: Right setting but wrong index. We don't need to put this setting on the set indices, as we generally access them by id.  However, in order to create a set, we need to query another index to determine which ids to put in the set. In the data portal, this is done by building a filter, then creating a set based on the filter.   

When we look at the graphql set creation classes, we can see the `doc_type` that is used for the query.  For example, file sets are created using the [file doc type](https://github.com/NCI-GDC/gdcapi/blob/master/gdcapi/graphql_schemas/mutations/sets/create/__init__.py#L101).  This doc type is then used to [look up the index name](https://github.com/NCI-GDC/gdcapi/blob/cc35bd5eeb59a3f00bcc811f4b757b7bcdc97aa2/gdcapi/graphql_schemas/mutations/sets/create/__init__.py#L58) to query.  In this case, it'd map to the graph index for files, which uses [this settings file](https://github.com/NCI-GDC/gdc-models/blob/master/es-models/gdc_from_graph/file.mapping.yaml).

The doc type to index name for graph indices are [defined in tungsten](https://github.com/NCI-GDC/tungsten/blob/64d729627e6db8e1be06fc5cc6bea0b5674446d3/salt/srv/services/api/conf/gdcapi.wsgi#L23-L28).

philtom-ctds (approved):

kulgan (approved):

PR #166, opened 2021-04-08
Standardize dependency management in gdc-models

qiaouchicago:
https://jira.opensciencedatacloud.org/browse/DEV-278

https://jira.opensciencedatacloud.org/browse/DEV-677

- Specify relevant direct dependency ranges in setup.py
- Specify dev/build dependency ranges in dev-requirements.in
- Use pip-compile to generate requirements.txt and dev-requirements.txt


ProfOak (commented):
This seems ok to me. I see that both Jenkins and Travis status checks are passing but do we want Travis anymore? If not then we will need to remove the .travis yaml and remove the status checks for it.

qiaouchicago:
> This seems ok to me. I see that both Jenkins and Travis status checks are passing but do we want Travis anymore? If not then we will need to remove the .travis yaml and remove the status checks for it.

Beautiful icon. Idk. I did not remove Travis because this one does not depend on any internal repos. I think travis is usually faster than Jenkins, probably good to keep it as we can get result faster?

XukunXiangUChicago (approved):

ProfOak (approved):

PR #174, opened 2022-01-28
DEV-1007: Re-adding fields removed between 2.7.1 and 2.8.1

mpsolano:
Added several fields removed between 2.7.1 and 2.8.1 back into es-mapping. The removal of these properties had an unintended consequence of causing some portal graphql queries to fail when validating against the graphql schema. Specifically case and file mappings were updated along with their relevant descriptions.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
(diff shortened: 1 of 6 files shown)


kulgan (approved):

kulgan (approved):
Can you rebase to the latest develop, the es related changes are already in there

philtom-ctds (approved):

PR #176, opened 2022-07-13
DEV-1267: Add missing fields to case centric model

BenX00:
These changes add the missing clinical data to the case centric model.


ProfOak (approved):
Do we want to target develop for this?

PR #177, opened 2022-10-25
DEV-1215: update mappings for case/gene centric index

kulgan:
Updates the _source excludes mapping for both gene and case centric mappings. The update is to allow the respective indices to now include case_id and gene_id, symbol.

Adds tests that mimics this behavior for validation


es-models/gene_centric/gene_centric.mapping.yaml
```diff
@@ -2,7 +2,13 @@ _size:
...
 _source:
   excludes:
-  - case.*
+  - case.*.*
+  - case.consent_type
+  - case.d*
+  - case.i*
+  - case.p*
+  - case.s*
+  - case.*_*_*
```
mpsolano, line 11: Would it make sense to comment these (and possibly the excludes section in case_centric) to provide additional details?

tests/test_dsls.py
```diff
@@ -0,0 +1,127 @@
+"""This module tests the behavior of using excludes in elasticsearch mappings"""
```
mpsolano, line 1: Should this also contain test cases for gene_centric to verify case_id is being returned?

XukunXiangUChicago (approved):
lgtm

tests/test_dsls.py
kulgan: The test cases were testing the elasticsearch excludes behavior when used with wildcards. It was trying to answer the question, when I exclude using a wildcard of this form `case.*_*` does it behave as expected. I wasn't testing whether gene_id shows up in case centric or not (at least not directly). The specific test for whether the fields show up or not was done manually by building an index and checking

es-models/gene_centric/gene_centric.mapping.yaml
kulgan: Sadly pre-commit strips off all comments from the yaml file :(

mpsolano: That makes sense. Thank you!

tests/test_dsls.py
mpsolano: Sounds good, thank you!

mpsolano (approved):
Looks good!

PR #183, opened 2023-03-11
DEV-1752: Update case_centric mapping to include all graph_case properties

mpsolano:
The case_centric indices are being updated to include all properties from graph_case, effectively making them a superset of that index. This is in preparation for switching cohort filters to run against the case_centric indices (via /case_ssms) instead of the graph_case index (via /cases).

### DIFF
--- a/es-models/case_centric/case_centric.mapping.yaml
+++ b/es-models/case_centric/case_centric.mapping.yaml
@@ -6,6 +6,47 @@ _source:
   - gene.biotype
   - gene.is_cancer_gene_census
 properties:
+  aliquot_ids:
+    type: keyword
+  analyte_ids:
+    type: keyword
+  annotations:
+    properties:
+      annotation_id:
+        type: keyword
+      case_id:
+        type: keyword
+      case_submitter_id:
+        type: keyword
+      category:
+        type: keyword
+      classification:
+        type: keyword
+      created_datetime:
+        type: keyword
+      creator:
+        type: keyword
+      entity_id:
+        type: keyword
+      entity_submitter_id:
+        type: keyword
+      entity_type:
+        type: keyword
+      legacy_created_datetime:
+        type: keyword
+      legacy_updated_datetime:
+        type: keyword
+      notes:
+        type: keyword
+      state:
+        type: keyword
+      status:
+        type: keyword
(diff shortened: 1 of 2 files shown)


BenX00 (approved):
Looks good to me. Once you get this updated you'll want to update MI with the new version and then test it out w/ a run. There may be some fields that need to be configured in the [include_as_array](https://github.com/NCI-GDC/gdc-dev-guide/wiki/Mutation-Indexer#case-centric) configuration. If so, we'll want to update the default config file in tungsten (not a show stopper but a nice to have for everyone else.)

jiakf (approved):

es-models/case_centric/case_centric.mapping.yaml
```diff
@@ -6,6 +6,47 @@ _source:
   - gene.biotype
   - gene.is_cancer_gene_census
 properties:
+  aliquot_ids:
+    type: keyword
+  analyte_ids:
+    type: keyword
```
kulgan, line 12: Do we need the normalizer definitions found in the case mappings here as well?

mpsolano: I hadn't seen them added anywhere for our previous work so I didn't carry them over. @BenX00 Do you have any thoughts on this?

kulgan (approved):

es-models/case_centric/case_centric.mapping.yaml
mpsolano: Thank you for sending the Jira ticket associated with this functionality @kulgan: https://jira.opensciencedatacloud.org/browse/DAT-2302

From this I found the following: "The cDAVE feature mainly relies on the case_centric mutation index. As the mutation indexer pulls the case data from the gdc_from_graph index when building the mutation indices, normalizing casing in the graph index should also normalize the casing in the mutation indices without any additional changes required."

This leads me to believe we don't need to include these from an indexing standpoint. Offline you mentioned this may come into play at query time so we still need to determine if that will impact this change.

mpsolano: This was discussed offline and the normalizer is not needed in the case_centric properties.

PR #188, opened 2023-07-21
DEV 2007 drop py35

qiaouchicago:
pyupgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 env:
   - ES_VERSION="7.7.1"
 python:
-  - "3.5"
+  - "3.6"
   - "3.7"
 
 before_install:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@ language: python
 env:
   - ES_VERSION="7.7.1"
 python:
-  - "3.5"
```
wrkhenddher, line 6: 😱😱😱😱

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: python
...
+  - "3.6"
   - "3.7"
```
wrkhenddher, line 7: Could we go up higher? (3.8 or even 3.9)

wrkhenddher (approved):

.travis.yml
qiaouchicago: I checked and am waiting for this:
https://github.com/NCI-GDC/esbuild/blob/develop/.travis.yml

mpsolano (approved):
Looks good!

PR #189, opened 2023-07-24
DEV-1986: Update gene expression model & settings.

BenX00:
Restructure the gene expression model into a flat one each representing a single gene for a given case.

### DIFF
--- a/es-models/gene_expression/settings.yaml
+++ b/es-models/gene_expression/settings.yaml
@@ -20,8 +20,7 @@ analysis:
       min_gram: '1'
       side: front
       type: edge_ngram
-index.mapping.nested_fields.limit: 100
-index.mapping.nested_objects.limit: 100000000
-index.max_inner_result_window: 20000
-index.max_result_window: 100000000
-index.max_terms_count: 100000
+index:
+  max_inner_result_window: 20000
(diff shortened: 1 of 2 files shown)


stilesj-uchicago (approved):

es-models/gene_expression/settings.yaml
```diff
@@ -20,8 +20,7 @@ analysis:
       min_gram: '1'
       side: front
       type: edge_ngram
-index.mapping.nested_fields.limit: 100
-index.mapping.nested_objects.limit: 100000000
-index.max_inner_result_window: 20000
-index.max_result_window: 100000000
-index.max_terms_count: 100000
+index:
```
wrkhenddher, line 23: Not sure but jic, let's note in one of our tickets that we may need to change how we retrieve max values. I'm thinking of https://jira.opensciencedatacloud.org/browse/DEV-1957 during Hippocrates.

wrkhenddher (approved):

es-models/gene_expression/settings.yaml
BenX00: This is noted in the ticket for this work [DEV-1986](https://jira.opensciencedatacloud.org/browse/DEV-1986)

PR #191, opened 2023-09-13
DEV-2143: add mapping diff/merge functionality

ProfOak:
The functionality was added here and will be used by esbuild/gdcapi. For this to work the version of deepdiff (6.3.1+) needed for the `force=True` parameter also needed py37+.


qiaouchicago (approved):

setup.py
```diff
@@ -11,11 +10,9 @@
     install_requires=[
         "PyYAML>=3.11,<6",
         "elasticsearch>=7.0.0,<8.0.0",
+        "deepdiff~=6.5.0",
```
wrkhenddher, line 13: Consider adding `python_requires`.

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement

tests/test_mapping_utils.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+from gdcmodels import mapping_utils
+
```
wrkhenddher, line 3: What happens in this case?

```python
one = {"a": [1,2]}
two = {"a": [1,3]}
```

But I don't know if we have properties that are list-like.

wrkhenddher (approved):

setup.py
```diff
@@ -11,11 +10,9 @@
...
+        "mergedeep~=1.3.4",
```
qiaouchicago, line 14: please remove restrictions for libraries.

ProfOak: This is not just a library. It's used in [elasticsearch](https://github.com/NCI-GDC/tungsten/blob/develop/salt/pillar/services.sls#L147-L148) vms.

tests/test_mapping_utils.py
ProfOak: This is a good observation because it will join the two even if there are what we would consider duplicates. I think this behavior is fine because we don't have this situation in the [mapping files](https://github.com/NCI-GDC/gdc-models/blob/develop/es-models/gdc_from_graph).

PR #198, opened 2023-12-15
DEV 2298 different prefix for different branch

qqiao2024:


stilesj-uchicago (approved):

setup.cfg
```diff
@@ -0,0 +1,56 @@
+[metadata]
+name = gdcmodels
+description = "Stores and serves GDC data models defined in static YAML files.
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = NCI GDC
+author_email = gdc_dev_questions-aaaaae2lhsbell56tlvh3upgoq@cdis.slack.com
+license = Apache Software License 2.0
+license_file = LICENSE
+url = https://github.com/NCI-GDC/gdcdatamodel2
```
wrkhenddher: gdcdatamodel **2**?

setup.cfg
```diff
@@ -0,0 +1,56 @@
...
+keywords = python3 gdcdatamodels gdcdictionary plaster
+classifiers =
+    Intended Audience :: Developers
+    Programming Language :: Python
+    Programming Language :: Python :: 3.8
```
wrkhenddher, line 15: 3.7 too?

tox.ini
```diff
@@ -25,16 +25,18 @@ commands = python-codacy-coverage -r coverage.xml
 [testenv:publish]
 changedir =
 passenv =
-  TWINE_*
+    TWINE_*
+    CI_COMMIT_*
 skip_install=true
 deps =
-    setuptools_scm
+    setuptools_scm==8.0.4
```
wrkhenddher, line 32: do we really need to pin setuptools_scm?
If so, maybe we should pin it in pyproject.toml instead?

tox.ini
```diff
@@ -25,16 +25,18 @@ commands = python-codacy-coverage -r coverage.xml
...
+    setuptools_git_versioning<2
```
wrkhenddher, line 33: maybe we should pin it in pyproject.toml instead?

wrkhenddher (approved):

setup.cfg
qqiao2024: 3.7 is removed from this repo

tox.ini
qqiao2024: already in there, second line.

tox.ini
qqiao2024: already in there, second line.

PR #199, opened 2023-12-15
DEV-2377 add a digest to a set to know if the IDs match those in post…

stilesj-uchicago:
…gres w/o sending them all over the write. This is to avoid expensive write operations.


mpsolano (approved):
Looks good!

wrkhenddher (approved):

PR #203, opened 2024-02-13
DEV-2498: Update Tests

BenX00:
This updates the tests to better abstract away the parts of them that will have to change when the new model directory structuring is introduced with this story.

Updates:

- Move single-use test fixture logic to inside the test in order to allow for clearer understanding of what the test setup looks like and therefore what is being tested.
- Introduce constants file paths withing es-models.
- Switch to `importlib_resource` vs depreciated `pkg_resources`
- Mock `pkg_resources` with a temp dir vs passing a directory directly. (This functionality appears only in test and will be phased out in code change)

NOTE: `pytest-cov` was causing issues in gitlab. It couldn't resolve the version for some reason. I was able to get around this by opting to use coverage direct.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ exclude =
 dev =
     coverage
     elasticsearch-dsl~=7.4
-    importlib-resources
+    importlib-resources;python_version<'3.9'
     pytest
     typing-extensions
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,9 +42,11 @@ exclude =
 
 [options.extras_require]
 dev =
+    coverage
     elasticsearch-dsl~=7.4
-    pytest~=7.0
-    pytest-cov~=4.0
+    importlib-resources
+    pytest
+    typing-extensions
 
 
 [options.entry_points]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,23 +1,25 @@
 import os
 import pathlib
+import tempfile
 from typing import Callable, Iterator
-import pkg_resources
 
 import elasticsearch
+import pkg_resources
 import pytest
 
 
 @pytest.fixture
-def es_models(tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch) -> pathlib.Path:
+def es_models(monkeypatch: pytest.MonkeyPatch) -> Iterator[pathlib.Path]:
     """Creates a temporary es-models directory and ensures via patching that it is
     loaded by the resources library.
     """
-    models = tmp_path / "es-models"
 
-    models.mkdir()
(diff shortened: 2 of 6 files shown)


setup.cfg
```diff
@@ -42,9 +42,11 @@ exclude =
 
 [options.extras_require]
 dev =
+    coverage
     elasticsearch-dsl~=7.4
-    pytest~=7.0
-    pytest-cov~=4.0
+    importlib-resources
```
wrkhenddher: iirc, this is part of python 3.7+ and a backport before that.
Maybe add "; python<=x.y" as needed?

https://docs.python.org/3.7/library/importlib.html#module-importlib.resources

tests/conftest.py
```diff
@@ -1,213 +1,23 @@
...
-    mapps = {
-        'foo': {
-            'mapping': foo_mapping,
-            'dir': foo_dir_name,
-            'descriptions': descriptions,
-        }
-    }
-    root = tmp_path
-    return mock_mappings(root, mapps)
+    models.mkdir()
```
wrkhenddher: never removed?

Could `with tempfile... as t:` help?

https://docs.python.org/3.8/library/tempfile.html?highlight=tempfile#module-tempfile

wrkhenddher (approved):

ProfOak (approved):

stilesj-uchicago (approved):

setup.cfg
BenX00: It is a backport but now for more than just 3.7. It now has the updated [interface](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files) introduced in 3.9 as well as some smaller tweaks from following releases; the old [interface](https://docs.python.org/3/library/importlib.resources.html#deprecated-functions) is now marked as depreciated and hence I like to just stick to the new one. 

So I can limit it to 3.9 but it is minorly annoying because we will now have to check the python version for every import vs the backport which is perfectly functional for all currently supported python versions.

```python
if sys.version_info < (3, 9):
    import importlib_resources as resources
else:
    from importlib import resources
```

tests/conftest.py
BenX00: I was just using the pytest fixture tmp_path that had already been in use w/ the assumption that it is cleaned up like tempfile. But you are correct that pytest does not clean these files up. I'll update things to use tmpfile.

setup.cfg
BenX00: @wrkhenddher any thoughts on how you'd like me to move forward? I don't want to merge this with this pending as it affects the review that follows too.

wrkhenddher: Nice! all good!

tests/conftest.py
wrkhenddher: Thanks!

PR #204, opened 2024-02-14
Feat/dev 2498 add blank vestigial properties

BenX00:
These changes create a new structure from which we load the gdc-models. Within this change:
- Added empty vestigial files to contain props that have been removed from the dict but are required in order not to break API functionality.
- Created logic to load these vestigial props when loading the models.
- Added type hints for the models that are loaded. 
- Removed redundant tags within the descriptions.yaml as logically they should only contain descriptions and not be generic meta files. 
- Simplified all mapping file names b/c their associated doc_type can now always be determined by their parent folder.


src/gdcmodels/__init__.py
```diff
@@ -1,62 +1,239 @@
...
+
+    Args:
+        detail: The detail with the relavant paths from which to load the mapping.
+        vestigial_included: A flag which determins if the vestial properties should be
+            included when loading the mapping.
+
+    Returns:
+        The ESMapping loaded from the paths within the given detail.
+    """
+    mapping = yaml.safe_load(detail.mapping.read_bytes())
```
ProfOak: Not super important but should we use this opportunity to use the C loaders? They are a bit faster.
```suggestion
    mapping = yaml.load(detail.mapping.read_bytes(), Loader=yaml.CSafeLoader)
```

ProfOak (approved):

src/gdcmodels/__init__.py
```diff
@@ -1,62 +1,239 @@
...
+
+
+Properties = Dict[str, Union["Property", "Autocomplete"]]
+
+
+class Property(TypedDict):
+    """A property in an es mapping."""
+
+    normalizer: NotRequired[str]
+    type: NotRequired[Literal["boolean", "double", "keyword", "long"]]
```
stilesj-uchicago, line 46: Why `NotRequired` instead of `Optional` ?

src/gdcmodels/es-models/case_centric/vestigial.yaml
```diff
@@ -0,0 +1 @@
+{}
```
stilesj-uchicago, line 1: All these vestigial files are in preparation to be populated in another step? I may have missed it, but I didn't see any new data for these.

stilesj-uchicago (approved):

src/gdcmodels/__init__.py
BenX00: `NotRequired` is a feature available only in typed dicts. It means the key may not be in the dictionary vs optional which would mean the key is always there but it's value may be `None`. Hence, we use them here as these structs are quite fluid as far as what may or may not be present in the dict. [Here](https://docs.python.org/3/library/typing.html#typing.NotRequired) are the docs for `NotRequired`.

src/gdcmodels/es-models/case_centric/vestigial.yaml
BenX00: These will be populated for the mappings that require them in the next step as far as current state (ie the difference between these models and what is used by esbuild/the actual indices in es.) A further step will be used to update the mapping with all new data from the graph and there may or may not be more introduce in this step as well

PR #206, opened 2024-03-04
DEV-2420: Standardize Pre-commit.

BenX00:
This standardizes our use of pre-commit with how it was being used in gdc-models. Hence, the standard set of pre-commits have been applied to all files & the custom yaml formatting has been removed as it is no longer in use.


kulgan (approved):
lgtm

stilesj-uchicago (approved):

PR #207, opened 2024-03-04
DEV-2420: Move logic in init into sub-modules.

BenX00:
The coming changes need functionality from the main modules `__init__` file. This ultimately results in circular dependencies. To insure this does not happen, the functionality in the `__init__` file has been separated and placed into sub-modules within the main one.


src/gdcmodels/utils.py
```diff
@@ -0,0 +1,15 @@
+from typing import IO, Any, Union
+
+import yaml
+
+
+def load_yaml(data: Union[str, bytes, IO[str], IO[bytes]]) -> Any:
```
kulgan, line 6: Usage of this function is quite similar to using `yaml.load` directly. I find using `yaml.load` directly more intuitive than loading it from a `utils`.

To use the utils
```python
from gdcmodels import utils

...
utils.load_yaml(...)
```

To use pyyaml
```python
import yaml

----
yaml.load(...)
```

src/gdcmodels/esmodels/__init__.py
```diff
@@ -0,0 +1,86 @@
+from typing import Any, Dict, KeysView, Literal, Mapping, Sequence, Union, overload
```
kulgan: nit: can the contents of this module be moved to a separate module that is not an `__init__` file

stilesj-uchicago (approved):
What was the driver for renaming es-models to esmodels?

BenX00:
> What was the driver for renaming es-models to esmodels?

`es-models` is an invalid module name. So it either has to be either `esmodels` or `es_models`. I chose the more abbreviated.

src/gdcmodels/utils.py
BenX00: I actually really prefer the utils approach. It makes it much easier to update the loader if we need to do so vs having to update every instance in which we are loading.

PR #208, opened 2024-03-04
DEV-2420: Port model mapper from normalizer.

BenX00:
The model mapper will become our **_standard_** way of interacting with the models. This will assume the all of the models will have settings and mappings which have been normalized & defaulted (logic to follow).


src/gdcmodels/mapper.py
```diff
@@ -0,0 +1,110 @@
+import re
```
stilesj-uchicago, line 3: Would you please add the moddoc here to explain the context of the module and why it is used?

src/gdcmodels/mapper.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import Any, Callable, Iterable, Iterator, Mapping, Optional, Tuple, Union
+
+import more_itertools
+
+from gdcmodels import esmodels
+
+Selector = Union[Callable[[Iterable[str]], Iterable[str]], str]
```
stilesj-uchicago, line 13: Can you add a comment to this structure? It is opaque to me.

stilesj-uchicago (approved):

src/gdcmodels/mapper.py
BenX00: I added a redirect. But don't want to add separate documentation as that will just get out of sync over time.

src/gdcmodels/mapper.py
```diff
@@ -0,0 +1,119 @@
...
+    def select_mapping(
+        self,
+        doc_type: str,
+        selector: Optional[Selector] = None,
+    ) -> Union[esmodels.ESMapping, esmodels.Property]:
+        """Select a sub-mapping or the entire mapping based on the doc_type.
+
+        Using the given doc_type any property within this mapping with that name will
+        have it's property definition returned. If multiple exist a selector should be
+        insure that the selection is deterministic.
```
kulgan:
```suggestion
        have it's property definition returned. If multiple exist a selector should
        insure that the selection is deterministic.
```

tests/test_mapper.py
```diff
@@ -0,0 +1,66 @@
...
+if sys.version_info < (3, 9):
+    import importlib_resources as resources
+else:
+    from importlib import resources
+
+
+@pytest.fixture
+def mock_mapper():
+    mapping_file = resources.files("tests") / "data/mapper/mapping.yaml"
+    mapping = utils.load_yaml(mapping_file.read_bytes())
```
kulgan: Since we are keeping the `utils` feature, you should consider making it more specific.`utils` easily become a dumping ground for things that do not fit in anywhere else and based on the current usage you want a customized yaml loader. This can be put into a more specific module or class rather than a generic `utils` module.

src/gdcmodels/mapper.py
```diff
@@ -0,0 +1,119 @@
...
+            selector: The selector can either be a callable which takes in all possible
+                paths to the given doc type which it filters to the desired path or the
+                name of a parent property that must be within the desired properties
+                path.
+
+        Returns:
+            The entire mapping definition or the property details associated with the
+            given doc_type.
+        """
+        if self.doc_type in [doc_type, f"{doc_type}_centric"]:
```
kulgan, line 104: Trying to understand why we have the extra `f"{doc_type}_centric"` item in the list. If `self.doc_type = case_centric`, for a particular instance passing a `doc_type = case` to this method returns the mapping for the `case_centric`. Is this the intended behavior?

src/gdcmodels/mapper.py
```diff
@@ -0,0 +1,119 @@
...
+            return self.mappings
+
+        mappings = {p: m for p, m in _walk_mapping(self.mappings) if p.endswith(f".{doc_type}")}
```
kulgan, line 107: Not related to this PR: We might need to look at alternative ways of identifying indices rather do logic based on whether or not a particular string appears in a field name or property. It really makes it hard to read and figure out if implementations will work correctly.

kulgan (approved):

src/gdcmodels/mapper.py
BenX00: For that you'd have to look to [here](https://github.com/NCI-GDC/gdc-mutation-indexer/blob/026db7739b0dd59dafdfbc40384880abe9e35e7e/src/mutation_indexer/builders/cnv_centric.py#L58) and [here](https://github.com/NCI-GDC/gdc-mutation-indexer/blob/026db7739b0dd59dafdfbc40384880abe9e35e7e/src/mutation_indexer/builders/ssm_centric.py#L60-L62). Unfortunately, the entire [model mapper](https://github.com/NCI-GDC/normalizer/blob/382d9cf328d098f0c889c04d9d249251f58c7942/normalizer/mapper.py#L146) is actually a structure that has its roots in MI. This logic is ultimately used to dynamically build a bunch of stuff in the [df_builders](https://github.com/NCI-GDC/gdc-mutation-indexer/blob/develop/src/mutation_indexer/builders/df_builders.py) via the [utils](https://github.com/NCI-GDC/gdc-mutation-indexer/blob/026db7739b0dd59dafdfbc40384880abe9e35e7e/src/mutation_indexer/builders/utils.py#L100-L111) module which, as you most certainly pointed out in your other comment, can be hard to read/understand/know what is coming out of those functions and know that they are working as intended. To add insult to injury, those builders are also SUPER inefficient compared to the actual builders in MI as they each rebuild basically the same data multiple times. 

These things all said for the time being we have to maintain this legacy interface until I (or someone else) gets the cycles to go switch the logic in MI to behave in a more predictable/stable/efficient way.

src/gdcmodels/mapper.py
BenX00: Tried to address this in my other comment as well cuz it's ultimately tied all together :anguished:

PR #209, opened 2024-03-05
Dev-2420: Port normalizer logic & standardize models

BenX00:
This adds the export process to the indices. This effort is to insure that the models loaded by `get_es_models` are the *_final_* models & that no more processes are transforming them before use. This PR only includes the work for the `viz` indices as the `graph` indices will follow. For the viz indices we:
- Add any default mapping values
- Add any default settings values
- Insure clinical_normalizer is applied to the appropriate fields.

NOTE: This includes the initial run of the process as the tests will now check that the models have been properly exported in case any manual updates were made to them.


mpsolano (approved):
Looks good!

src/gdcmodels/export/__main__.py
```diff
@@ -0,0 +1,4 @@
+from gdcmodels.export import cli
```
kulgan: Can this be added to the entrypoints in setup.cfg? There is already a script there but not sure if its still in use

README.md
```diff
@@ -17,6 +17,30 @@ Git repository centrally stores and serves GDC data models defined in static YAM
 
 ## Update the data models
 
+### Export
+
+Exporting is the process of updating the models with any properties which may be derived from external sources, normalizing keywords, as well as insuring all default mapping values are set. This process should be run after the gdcdictionary is updated and when any new property is added to the viz indices.
+
+The process can be run for any index (-i) and any of its doc-types (-d). Multiple can be spcified in the command line and if none are provided for either all of the respective type are run.
```
stilesj-uchicago:
```suggestion
The process can be run for any index (-i) and any of its doc-types (-d). Multiple can be specified on the command line and if none are provided for either all of the respective type are run.
```

README.md
```diff
@@ -17,6 +17,30 @@ Git repository centrally stores and serves GDC data models defined in static YAM
...
+Run all associated doc-types:
+```bash
+python -m gdcmodels.export -i gdc_from_graph -i case_centric
+```
+
+Run a singular doc-type:
+```bash
+python -m gdcmodels.export -i gdc_from_graph -d file
+```
+
```
stilesj-uchicago, line 44: Can you extend this section as to what to do after the export?

src/gdcmodels/esmodels/case_centric/settings.yaml
```diff
@@ -31,3 +37,6 @@ index:
       limit: 2000
   max_result_window: 100000000
   max_terms_count: 100000
+  number_of_replicas: 0
+  number_of_shards: 12
```
stilesj-uchicago, line 41: Hmm... I'm unsure if this is the best spot for the index settings. It does have good locality with the mappings. I don't have a suggestion but it seems like we're mixing configuration with data so it is setting off my spidey-sense.

setup.cfg
```diff
@@ -47,6 +47,8 @@ dev =
     coverage
     elasticsearch-dsl~=7.4
     pytest
+export =
```
stilesj-uchicago: Is the scenario for this `extra` that sometimes we want to export the ES configuration for ES build as a standalone process, but most of the time the gdc-models are used like a library like in the gdcapi?

src/gdcmodels/export/cli.py
```diff
@@ -0,0 +1,164 @@
+"""The basic cli functionality for the export."""
```
stilesj-uchicago: Would you please add the context on when this is needed to be run?

src/gdcmodels/export/cli.py
```diff
@@ -0,0 +1,164 @@
...
+    _ = old_mapping.pop("_meta", None)
+
+    exporter = EXPORTERS[index_name][doc_type]
+    new_mapping, new_settings = exporter.export(old_mapping, old_settings)
+    descriptions = new_mapping.pop("_meta", {}).get("descriptions")
+
+    diff = deepdiff.DeepDiff(new_mapping, old_mapping)
+    delta = VestigialDelta(diff, serializer=extraction_utils.dump_yaml)
+
+    assert (new_mapping + delta) == old_mapping
```
stilesj-uchicago: Should the `new_mapping + delta` _contain_ the `old_mapping` or exactly match it?

src/gdcmodels/esmodels/models.py
```diff
@@ -2,6 +2,19 @@
 
 from typing_extensions import NotRequired, Protocol, TypedDict
 
+GRAPH_INDICES = tuple(("gdc_from_graph", d) for d in ("annotation", "case", "file", "project"))
+VIZ_INDICES = tuple(
+    (i, i)
```
stilesj-uchicago, line 7: Is the VIZ_INDICES tuple structure so that it kind of matches the footprint of the GRAPH_INDICES structure?

stilesj-uchicago (approved):

README.md
```diff
@@ -17,6 +17,30 @@ Git repository centrally stores and serves GDC data models defined in static YAM
 
 ## Update the data models
 
+### Export
```
kulgan: nit: I'm thinking update or sync-models aligns more with the definition than export. Might be graphmanager's export feature messing with my head :)

kulgan (approved):

setup.cfg
BenX00: Yup. We only want to install/run export when we are working within the repo currently. When we install the library elsewhere we are just trying to access the models as they are

src/gdcmodels/export/cli.py
BenX00: It should be an exact match. Although we only _write_ the new fields that were added, this sanity check will apply the full delta including removing the newly created fields.

src/gdcmodels/export/cli.py
BenX00: I tend to not favor duplicating documentation as then it just becomes out of sync. Also I tried to just detail the general purpose of the module here and [here](https://github.com/NCI-GDC/gdc-models/blob/1d3f291309651cbbcce0a4c9e98fefb129f16f06/README.md#examples) in the README, I've included the when and how to run the export.

src/gdcmodels/esmodels/models.py
BenX00: Yes and that is the current state of things. It lends to a standard modeling of the results of `get_es_models` as well i.e. `index/doc_type` which for the viz indices is just a repeat.

README.md
BenX00: I added an after section. It's quite straight forward: review/commit/merge.

src/gdcmodels/esmodels/case_centric/settings.yaml
BenX00: As it stands this is where the settings for the indices live (well here and normalizer). If we need to move them we can but that can be done as an extra story.

PR #210, opened 2024-03-05
DEV-2420: Export Graph.

BenX00:
These changes export the graph from the data model & apply all defaults as well as the clinical normalizer.


mpsolano (approved):
Looks good!

src/gdcmodels/esmodels/gdc_from_graph/case/vestigial.yaml
```diff
@@ -61,6 +61,12 @@ dictionary_item_added:
     type: double
   root['properties']['exposures']['properties']['height']:
     type: double
+  root['properties']['exposures']['properties']['marijuana_use_per_week']:
```
stilesj-uchicago, line 64: Are we going manually add vestigial entries as part of the dictionary release?

src/gdcmodels/esmodels/gdc_from_graph/descriptions.yaml
```diff
@@ -87,8 +87,8 @@ annotations.analyte.submitter_id: A project-specific identifier for a node. This
   of the uuid for identifying or recalling a node.
 annotations.analyte.updated_datetime: A combination of date and time of day in the
   form [-]CCYY-MM-DDThh:mm:ss[Z|(+|-)hh:mm]
-annotations.analyte.well_number: Numeric value that represents the the well location
-  within a plate for the analyte or aliquot from the sample.
+annotations.analyte.well_number: Numeric value that represents the well location within
```
stilesj-uchicago, line 90: I'm going to assume these additions are all correct. The ole gdc-models have fallen behind a bit.

src/gdcmodels/export/graph/case.py
```diff
@@ -0,0 +1,65 @@
...
+                **self._cases.export_properties(),
+                "files": esmodels.Property(properties=self._get_file_properties(), type="nested"),
+            }
+        }
+
+        self._add_autocomplete(
+            mapping,
+            "case_autocomplete",
+            (
+                "case_id",
```
stilesj-uchicago: Looks like a candidate for external configuration.

stilesj-uchicago (approved):

src/gdcmodels/export/graph/case.py
BenX00: Not sure I'm following. Where externally would you want this configured?

stilesj-uchicago: I'm trying to think through what a developer would be doing when they update the models. They are going to update the mapping and vestigial.yaml files. Would it make sense to drop in an autocomplete.yaml next to those?

BenX00: For the graph indices, we shouldn't be touching the mapping nor the vestigial yaml files. They are completely generated from scratch in this script as, at least the mappings, were generated in esbuild (before we "broke" things by deleting fields.) 

If you look at the sync process [here](https://github.com/NCI-GDC/gdc-models/blob/8bb9557017573d202d4e11505a3514c7a1b5c7f2/src/gdcmodels/sync/graph/common.py#L278-L282), you'll see that for the graph that it just resets the mappings completely to what is generated. So if a dev were to try and manually update a mapping field that would not be generated by the graph sync logic:
A) if they didn't then run the sync process, [this](https://github.com/NCI-GDC/gdc-models/blob/8bb9557017573d202d4e11505a3514c7a1b5c7f2/tests/test_sync.py#L23-L36) test would fail as it insures that the synced mappings are the same as those loaded by `get_es_models`.
B) their newly added fields would instantly appear in the vestigial properties after sync.

src/gdcmodels/esmodels/gdc_from_graph/descriptions.yaml
BenX00: You can def double check the data models to be sure. I did spot check but that obviously isn't 100%. E.g. specimen_type is a field that was added after we stopped updating this file with the output of esbuild. Other places I mainly see typo fixes etc.

src/gdcmodels/esmodels/gdc_from_graph/case/vestigial.yaml
BenX00: If by manually, you mean we'll need to run the sync script w/ the updated models. Then that is correct. But we should NOT be manually editing these files. That should be left to the sync script.

stilesj-uchicago: Great! Because I wanted no part of manually editing anything!

src/gdcmodels/export/graph/case.py
stilesj-uchicago: I'm glad to hear that mapping and vestigal.yaml files are generated. The 'auto_complete' configuration still needs to be manually editing in the source code. Is there any indicators in the dictionary to drive this value? Or is it unique for each?

BenX00: As far as I know, the business drives what is included in the autocomplete feature.

PR #211, opened 2024-03-07
DEV-2420: Standardize Loading of Models

BenX00:
This is the final phase of the standardization process. This forces anyone who wishes the user to load the models to do so through this singular point of entry and to interact with the FINAL mappings/settings through the model mapper objects. No modification should be done to the models after they are loaded as they should reflect the state of all mappings used in the system.


mpsolano (approved):
Looks good!

stilesj-uchicago (approved):

PR #215, opened 2024-04-23
DEV-2655: Create aliases with saved set indices

ProfOak:
Relates to: https://github.com/NCI-GDC/tungsten/pull/2400

### DIFF
--- a/src/gdcmodels/init_index.py
+++ b/src/gdcmodels/init_index.py
@@ -2,7 +2,7 @@
 
 import argparse
 import sys
-from typing import cast
+from typing import List, cast
 
 import elasticsearch
 from typing_extensions import Iterable, Optional, Protocol
@@ -11,7 +11,8 @@ import gdcmodels
 
 
 class Arguments(Protocol):
-    index: str
+    index: List[str]
+    alias: List[str]
     prefix: str
     host: str
     port: int
@@ -37,6 +38,12 @@ def get_parser() -> ArgumentParser:
         required=True,
         nargs="+",
         help="index to be initialized",
+    )
+    parser.add_argument(
+        "--alias",
+        dest="alias",
+        nargs="+",
+        help="aliases to be initialized",
     )
     parser.add_argument(
         "--prefix", dest="prefix", required=True, help="prefix for the index name"
@@ -119,11 +126,20 @@ def init_index(args: Arguments):
     es_models = gdcmodels.get_es_models(vestigial_included=False)
     es = get_elasticsearch(args)
 
-    for index in args.index:
+    aliases = ["" for _ in args.index]
+    if args.alias:
+        aliases = args.alias[:]
+
+    # You cannot create a list of indices and provide a partial list of aliases.
+    if len(args.index) != len(aliases):
+        print(f"Mismatching arguments for index: {args.index} and alias: {args.alias}")
+
+    for index, alias in zip(args.index, aliases):
         if not es_models.get(index):
             print(f"Specified index '{index}' is not defined in es-models, skipping it!")
             continue
 
+        indices_created = 0
         for index_type in es_models[index].keys():
             if index_type == "_settings":
                 continue  # settings, not index type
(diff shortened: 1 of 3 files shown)


src/gdcmodels/init_index.py
```diff
@@ -119,11 +126,20 @@ def init_index(args: Arguments):
     es_models = gdcmodels.get_es_models(vestigial_included=False)
     es = get_elasticsearch(args)
 
-    for index in args.index:
+    aliases = ["" for _ in args.index]
```
stilesj-uchicago: Can you comment on this construct? It is initializing a list with empty strings to match the indices. That smells like a map or named tuples. Beware keeping two lists of synced items.

src/gdcmodels/init_index.py
```diff
@@ -157,6 +173,33 @@ def init_index(args: Arguments):
...
+
+        # This script uses the index names as defined by es_models and creates
+        # indices from those instead of naively creating elements from the
+        # --index flag directly. That means 'gdc_from_graph' can create
+        # multiple indices despite only being one of the elements in the
+        # --index flag. This breaks the alias logic but we don't use this
+        # script to create those indices so this conditional should never
+        # happen in practice.
+        if indices_created > 1:
+            print(f"Cannot create alias '{alias}' because too many indices created for '{index}'")
```
stilesj-uchicago: `print`? Should this be logging instead?

src/gdcmodels/init_index.py
```diff
@@ -38,6 +39,12 @@ def get_parser() -> ArgumentParser:
         nargs="+",
         help="index to be initialized",
     )
+    parser.add_argument(
```
stilesj-uchicago, line 66: Food for thought, an index has an alias attribute. Aliases are not a first order structure in ES. If an index is removed that alias is also removed. Does this account for indices that share the same alias?

stilesj-uchicago (approved):

src/gdcmodels/init_index.py
ProfOak: I'm not sure if I'm understanding correctly but if the indices are removed the aliases also go with them.

src/gdcmodels/init_index.py
ProfOak: I typically follow conventions in the code base I'm working on but it sounds like maybe I should have ripped out any existing logging and converted it to use the python logger instead. I can make those changes.

src/gdcmodels/init_index.py
stilesj-uchicago: If the last index using an alias is removed, then the alias no longer available.

src/gdcmodels/init_index.py
```diff
@@ -119,44 +142,87 @@ def init_index(args: Arguments):
...
             )
+            continue
+
+        if es.indices.exists_alias(name=index_builder.alias_name):
+            logger.warning(
+                f"Alias '{index_builder.alias_name}' exists already, skipping creating."
+            )
+            continue
+
+        es.indices.put_alias(name=index_builder.alias_name, index=index_builder.full_index_name)
```
mpsolano, line 226: Since you have a logger.info statement for creating the index, would it make sense to also have a logger.info statement for creating the alias?

src/gdcmodels/init_index.py
```diff
@@ -1,17 +1,42 @@
...
 import elasticsearch
 from typing_extensions import Iterable, Optional, Protocol
 
 import gdcmodels
 
+logger = logging.getLogger("init_index")
+
+
+@dataclasses.dataclass
+class ESIndexBuilder:
```
mpsolano, line 18: Thank you Jimmy for clarifying offline the reason why different logic is used for the non set indices aliasing as compared to the set indices aliasing. Specifically, you pointed out that there is a separate script that exists for creating the set indices which does not have access to the aliasing logic in the non-set indices script. This can be seen on steps 3c and 3d in the dev cluster sop: https://gdc-ctds.atlassian.net/wiki/spaces/GDC/pages/63897608/GDC+Cluster+SOP#:~:text=Now%20run%20the%20other%20script%20to%20create%20the%20saved%20sets%20indices%20(this%20should%20also%20be%20quick%20to%20run)%3A

mpsolano (approved):
Looks good!

PR #217, opened 2024-05-07
DEV-2677: Add curated model index to excluded fields.

BenX00:
This change ensures that the new `curated_model_index` is not included in the generated mappings. It slightly refactors the code so logic calling the `_load_properties_from` function can control what fields are ignored.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,8 +49,8 @@ dev =
     pytest
 sync =
     click
-    gdcdictionary
-    gdcdatamodel2
+    gdcdictionary~=3.0.0
+    gdcdatamodel2~=3.0.0
 
 [options.entry_points]
 console_scripts =
--- a/src/gdcmodels/sync/graph/common.py
+++ b/src/gdcmodels/sync/graph/common.py
@@ -1,6 +1,6 @@
 import abc
 import functools
-from typing import Any, Container, DefaultDict, Iterable, Mapping, Optional, TypeVar, cast
+from typing import Any, Container, DefaultDict, Iterable, Mapping, Optional, Tuple, TypeVar, cast
 
 import gdcdictionary
 from gdcdatamodel2 import models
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -49,8 +49,8 @@ dev =
     pytest
 sync =
     click
-    gdcdictionary
-    gdcdatamodel2
+    gdcdictionary~=3.0.0
```
stilesj-uchicago, line 52: Why pin? Were you getting 3.1.0 artifacts in the requirements.txt?

src/gdcmodels/sync/graph/common.py
```diff
@@ -273,39 +273,45 @@ def _get_nodes_by_category(category: str) -> Iterable[models.Node]:
...
+    def is_included(property: Tuple[str, Container[type]]) -> bool:
+        return property[0] not in excluded_fields
+
     properties = NestedDict()
 
     if include_id:
         properties[f"{node.get_label()}_id"] = ESProperty.string()
 
-    for field, types in node.get_pg_properties().items():
+    for field, types in filter(is_included, node.get_pg_properties().items()):
```
mpsolano, line 291: 👍

mpsolano (approved):
Looks good!

setup.cfg
BenX00: I'm pinning to keep this PR from being about the actual upgrade of the mappings. That will be part of work to come.

qqiao2024 (approved):
~The gdcdictionary and gdcdatamodel2 should be in same version~
we probably do not need gdcdictionary.

setup.cfg
qqiao2024: Did a quick search and I do not think we need gdcdictionary at all

BenX00: Not sure where you were searching but the sync process uses gdcdictionary to load the descriptions for the properties [here](https://github.com/NCI-GDC/gdc-models/blob/0b000e9ed2532a2d3b41f2c2543f1ee7ce3d544a/src/gdcmodels/sync/graph/common.py#L102-L119).

stilesj-uchicago (approved):

PR #219, opened 2024-05-30
Dev-2749: Add Other Clinical Attributes

BenX00:


src/gdcmodels/esmodels/case_centric/mapping.yaml
```diff
@@ -2416,6 +2416,173 @@ properties:
         type: nested
       nadir_cd4_count:
         type: double
+      other_clinical_attributes:
```
stilesj-uchicago, line 2419: How did you go about populating these segments?

stilesj-uchicago (approved):

src/gdcmodels/esmodels/case_centric/mapping.yaml
BenX00: [This](https://github.com/NCI-GDC/gdc-models/blob/e407b50d4eb7ef8e8f0180fc4188042ca514a938/src/gdcmodels/sync/graph/common.py#L476-L478) populates the new structure in the `graph/case` & `graph/file` mappings' follow_up struct. The `case_centric` just copies whatever is found in the `graph/case` mapping.

qqiao2024 (approved):

PR #221, opened 2024-09-05
DEV-2968: Sync Models w/ Dictionary

BenX00:
This syncs models with the dictionary. It also updated the tests to be deterministic by insuring the same versions of the dictionary and data model are used in the tests.


requirements-sync.txt
```diff
@@ -0,0 +1,79 @@
+#
```
stilesj-uchicago, line 1: It took me a second to piece together what is going on here. Would you please update README.md about installing using this file to run the sync tool as this is a little non-standard? Maybe add a comment on how this repository contains both a library and a tool.

stilesj-uchicago (approved):

jiakf (approved):

requirements-sync.txt
BenX00: So there is a description of this process the [readme](https://github.com/NCI-GDC/gdc-models?tab=readme-ov-file#update-the-data-models) and there are more docs for the models [here](https://gdc-ctds.atlassian.net/wiki/spaces/GDC/pages/347570178/GDC+Models). I've updated the readme with the updated installation cmds but otherwise have left it as is. If you think further docs are needed, let me know or feel free to extend the confluence as well.

PR #229, opened 2024-12-17
DEV-3186: add /segment_cnv_centric mapping

jiakf:


src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
```diff
@@ -0,0 +1,606 @@
...
+              tumor_grade:
+                normalizer: clinical_normalizer
+                type: keyword
+              year_of_diagnosis:
+                type: long
+            type: nested
+          disease_type:
+            normalizer: clinical_normalizer
+            type: keyword
+          exposures:
```
mpsolano, line 416: The `cnv_occurrence_centric` mapping includes `bmi` under exposures. Since the ticket specifies this should be identical, should it either be added here or perhaps removed from `cnv_occurrence_centric`?

src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
```diff
@@ -0,0 +1,606 @@
+---
+_size:
+  enabled: true
+dynamic: strict
+properties:
+  chromosome:
```
mpsolano, line 6: The CNV fields for `cnv_occurrence_centric` are nested under a `cnv` property whereas in this mapping they appear at the root of the document. I just want to verify this is the desired shape.

mpsolano (approved):
Looks good!

src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
jiakf: oh the way i put it in the jira ticket is not clear. it says the cnv_centric "occurrence" object, not the "cnv_centric_occurrence" index

src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
jiakf: same comment as above

src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
jiakf: https://gdc-ctds.atlassian.net/browse/DICT-170

so bmi under exposure is being deprecated, so I'm thinking 1) we don't need it for this mapping under exposures 2) it probably should be removed in cnv_occurrence_centric

mpsolano: Ah, that's my bad, I misread. Thank you for the clarification. And agreed, no need to include bmi here. We should probably remove bmi from cnv_occurrence_centric but that's out of scope for this effort.

src/gdcmodels/esmodels/segment_cnv_centric/mapping.yaml
mpsolano: Thank you for clarifying

stilesj-uchicago (approved):
I would like to see module doc (or in the README) in the esmodels/__init__.py describing the folder structure and purpose of the 3 file types found in each folder. I think that would be a nice breadcrumb to leave to maintainers.

PR #236, opened 2025-08-15
DEV-3637: Update sync logic for viz indices.

BenX00:


src/gdcmodels/sync/viz.py
```diff
@@ -1,93 +1,419 @@
...
 
 import gdcmodels
 from gdcmodels import esmodels
 from gdcmodels.sync import common
 
+UNNORMALIZED_FIELDS = (
+    common.DefaultNormalizerSynchronizer.DEFAULT_EXCLUDED_PROPERTIES
+    - frozenset(("case_submitter_id", "entity_submitter_id"))
+)
+"""Fields with in the mappings which should NOT have the clinical normalizer applied."""
```
mpsolano: Typo in the comment?

src/gdcmodels/sync/viz.py
```diff
@@ -1,93 +1,419 @@
...
+NORMALIZER_SYNCHRONIZER = common.DefaultNormalizerSynchronizer(
+    excluded_properties=UNNORMALIZED_FIELDS
+)
+"""The synchronizer for add the clinical normalizer to the appropriate fields."""
```
mpsolano, line 24: I don't see a lot of docstring comments associated with constants. Is this something typically done? If not, I might recommend switching to regular comment.

mpsolano (approved):

src/gdcmodels/sync/viz.py
BenX00: These actually appear in type hints at least in VS code. So I'd advocate doing them as such.
image

src/gdcmodels/sync/viz.py
BenX00: I assume you just mean the "within" bit?

mpsolano: I actually just misread the comment, I didn't even notice that. But looks good, thank you!

src/gdcmodels/sync/viz.py
mpsolano: Oh nice! Alright, I'm onboard, I think it makes sense to do this going forward. Thank you Micky

PR #241, opened 2026-05-28
DEV-4065: Create static overlay yamls.

BenX00:
These are all the statically mapped configs/properties. It should be noted that this is for current functionality which means the non-case_centric viz indices are **completely** static. The final word done here will be to move to a more dynamically driven overlay.


mpsolano (approved):

src/gdcmodels/sync/overlays/static/viz/segment_cnv_occurrence_centric.yaml
```diff
@@ -0,0 +1,245 @@
...
+          ann_arbor_pathologic_stage: *keyword
+          burkitt_lymphoma_clinical_variant: *keyword
+          cause_of_death: *keyword
+          classification_of_tumor: *keyword
+          cog_renal_stage: *keyword
+          colon_polyps_history: *keyword
+          days_to_diagnosis: *long
+          days_to_hiv_diagnosis: *long
+          days_to_last_follow_up: &double
+            type: double
```
Inarus899, line 52: I'm seeing this pattern, where many properties are labeled *double, but then there will be one that is labeled &double and then specified as type: double. Not sure if that is a problem, just a variation that seems off.

Inarus899 (approved):

src/gdcmodels/sync/overlays/static/viz/segment_cnv_occurrence_centric.yaml
BenX00: Those are anchors & aliases. They are used to reduce the amount of redundant blocks. [Here](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) is a pretty good summary.

PR #246, opened 2026-06-01
Dev-4065: Sync metadata for mappings.

BenX00:


src/gdcmodels/esmodels/case_centric/mapping.yaml
```diff
@@ -1,4 +1,2538 @@
...
+    - samples.annotations
+    - samples.portions
+    - samples.portions.analytes
+    - samples.portions.analytes.aliquots
+    - samples.portions.analytes.aliquots.annotations
+    - samples.portions.analytes.annotations
+    - samples.portions.annotations
+    - samples.portions.slides
+    - samples.portions.slides.annotations
+  descriptions:
```
mpsolano, line 43: Will this result in descriptions coming back in API for viz indices mappings where currently there are none? For reference, try the following endpoints in API that show the difference in descriptions between the cases endpoint (based on the graph_case index) and the case_ssms endpoint(based on the case_centric index):
`GET https://api.gdc.cancer.gov/cases/_mapping`
vs
`GET https://api.gdc.cancer.gov/case_ssms/_mapping`

To do that, gdcapi uses gdcmodels [here](https://github.com/NCI-GDC/gdcapi/blob/07bf087f3ed75051690600f9e9ff6000b5dd59af/src/gdcapi/graphql_schemas/mappings/__init__.py#L19).

To be clear, if so I think that's a good thing but just want to verify the behavior is intended and documented and something QA is prepared to test.

src/gdcmodels/esmodels/case_centric/mapping.yaml
```diff
@@ -1,4 +1,2538 @@
...
+      public, private, or parochial schools, and in colleges, universities, or professional
+      schools.
+    cases.demographic.ethnicity: An individual's self-described social and cultural
+      grouping, specifically whether an individual describes themselves as Hispanic
+      or Latino. The provided values are based on the categories defined by the U.S.
+      Office of Management and Business and used by the U.S. Census Bureau.
+    cases.demographic.gender: 'Text designations that identify gender. Gender is described
+      as the assemblage of properties that distinguish people on the basis of their
+      societal roles. [Explanatory Comment 1: Identification of gender is based upon
+      self-report and may come from a form, questionnaire, interview, etc.]'
```
Inarus899, line 107: why is this one quoted and not the others?

Inarus899 (approved):

src/gdcmodels/esmodels/case_centric/mapping.yaml
BenX00: That is a bit of the hope. That said if it does not work for the viz indices there is nothing lost either.

src/gdcmodels/esmodels/case_centric/mapping.yaml
BenX00: My guess is b/c it has a `:` in the value.

PR #250, opened 2026-08-06
DEV-4208 - Add min/max values to gdc-models _meta field

amlehman:


src/gdcmodels/sync/overlays/graph/structures.py
```diff
@@ -340,10 +360,50 @@ def _load_node_descriptions(self, root: str) -> Mapping[str, str]:
 
         return descriptions
 
+    def _search_details_for_constraint(self, field: str, details: Mapping[str, str]):
+        if details.get("oneOf"):
+            details_entry = [d[field] for d in details["oneOf"] if d.get(field) is not None]
+            if details_entry:
+                return details_entry[0]
```
BenX00: Just for future reference. It's pretty frowned upon to use this approach. In fact, I'm surprised ruff didn't yell at you. Basically you shouldn't create an entire list to just access the first item. In little cases like this it doesn't mean much but if the underlying data were to grow to an large size this approach can load a lot of garbage into memory all just to get the first element. Hence, we should always default to a more robust style which is optimal regardless. Hence, an iterator + next works i.e. `next(d for d in ...)` or even better more_itertools has a `first` function i.e. `more_itertools.first(d for d in ...)`. First also has options for a default value if something more meaningful than `None` as a return is desired. Further, `more_itertools.first_true` does all the same things as `first` + you can supply a `pred` which can be used to filter the input.

src/gdcmodels/sync/overlays/graph/structures.py
```diff
@@ -340,10 +360,50 @@ def _load_node_descriptions(self, root: str) -> Mapping[str, str]:
 
         return descriptions
 
+    def _search_details_for_constraint(self, field: str, details: Mapping[str, str]):
```
BenX00: It's always important to have the correct type hints: `Mapping[str, str]` should prob be `Mapping[str, Any]`. Your type checker should have been complaining about the code below @ `d.get(...)` as `get` is not a method on a string. AKA make sure you have a type checker set up. I always recommend [basedpyright](https://docs.basedpyright.com/latest/)

src/gdcmodels/sync/overlays/graph/structures.py
```diff
@@ -340,10 +360,50 @@ def _load_node_descriptions(self, root: str) -> Mapping[str, str]:
...
+
+        for prop, details in self._load_schema_properties():
+            field_constraints = {}
+            minimum = self._search_details_for_constraint("minimum", details)
+            if minimum:
+                field_constraints["minimum"] = minimum
+
+            maximum = self._search_details_for_constraint("maximum", details)
+            if maximum:
+                field_constraints["maximum"] = maximum
```
BenX00: This approach seems strange to me. I feel like the min/max should both be coming from the same source which is not guaranteed in the current implementation. That is to say, min could come from one "oneOf" and max another. That feels oddly inconsistent. 

```suggestion
            details = more_itertools.first_true(
                # Either iterate one of (if any) or just the root details
                details.get("oneOf") or (details,),
                # Find the first w/ any keys which match our desired
                # constraints.
                # MIN_MAX_CONSTRAINTS = frozenset({"minimum", "maximum"})
                pred=lambda d: d.keys() & MIN_MAX_CONSTRAINTS,
                # An empty default is always better than None :D
                default={},
            )
            field_constraints = {c: details[c] for c in MIN_MAX_CONSTRAINTS if c in details}
```

BenX00: If we really do want to cast the wide net:

```suggestion
            field_constraints = {
                constraint: subdetails[constraint]
                for constraint in MIN_MAX_CONSTRAINTS 
                for subdetails in itertools.chain((details,), details.get("oneOf", ()))
                if constraint in subdetails
            }
```

BenX00: If nothing else there is a bug when min or max are `0`. They won't be added to the resulting `field_constraints` as they also evaluate to a False.

src/gdcmodels/sync/overlays/graph/structures.py
amlehman: Is `Any` or `dict` more preferable here?

BenX00: dict is ideal if we know it's a dict.

src/gdcmodels/esmodels/case_centric/mapping.yaml
```diff
@@ -41,6 +41,618 @@ _meta:
...
+      minimum: -32872
+    cases.demographic.occupation_duration_years:
+      maximum: 89
+      minimum: 0
+    cases.demographic.year_of_birth:
+      minimum: 1900
+    cases.demographic.year_of_death:
+      minimum: 1900
+    cases.diagnoses.age_at_diagnosis:
+      maximum: 32872
```
stilesj-uchicago, line 68: What is the unit for this value? Days? If so, that caps the maximum age at just over 90yrs

stilesj-uchicago (commented):
Who can validate all the ranges are correct?

src/gdcmodels/esmodels/case_centric/mapping.yaml
amlehman: Yes, it's days. We use 365.25 as our days in a year value so it's actually just under 90 years. That's always been the max allowed age on the frontend, I'm not sure why it's capped at that @wteouchicago ?

wteouchicago: Yup, what Amy said. It's days and it's a magic number we use that makes it capped at just under 90 years for HIPAA Safe Harbor reasons (ages over 90 need to be obfuscated to reduce the chances of identifying a person). The conversion should be 365.25.

BenX00 (approved):

stilesj-uchicago (approved):

